PR #20, opened 2026-01-27
feat(provider): add Llamafile provider and OpenAI-compatible base package

agpituk:
**Summary**                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                             
  - Adds a reusable base provider in providers/openai/ for OpenAI-compatible APIs                                                                                                                                                                                            
  - Implements Llamafile provider as a thin wrapper using the base                                                                                                                                                                                                           
  - Makes the OpenAI provider itself a thin wrapper using the same base                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                             
  **Architecture**                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                             
  Base provider in `providers/openai/base.go`                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                             
  Many LLM services expose OpenAI-compatible APIs without providing an official Go SDK (Llamafile, vLLM, LM Studio, llama.cpp server, etc.). Rather than duplicating the OpenAI SDK integration code for each, this PR introduces a shared base provider within the openai   
  package.                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                             
  The base provider is configurable via BaseConfig:                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  ```
type BaseConfig struct {                                         
      APIKeyEnvVar   string                // Environment variable for API key                                                        
      BaseURLEnvVar  string                // Environment variable for base URL                                                       
      Capabilities   providers.Capabilities                        
      DefaultAPIKey  string                // Default key for local servers                                                           
      DefaultBaseURL string                // Default API endpoint 
      Name           string                // Provider name for error messages                                                        
      RequireAPIKey  bool                  // Whether API key is required                                                             
  }      
```                                                          

  **Llamafile provider**                                               

  Llamafile is a thin wrapper that configures the base:            

 ```
 import "github.com/mozilla-ai/any-llm-go/providers/openai"       

  type Provider struct {                                           
      *openai.BaseProvider                                         
  }         
```                                                       

 ```
 func New(opts ...config.Option) (*Provider, error) {             
      base, err := openai.NewBase(openai.BaseConfig{               
          Name:           "llamafile",                             
          DefaultBaseURL: "http://localhost:8080/v1",              
          // ...                                                   
      }, opts...)                                                  
      return &Provider{BaseProvider: base}, nil                    
  }      
```                                                          

  When to use base vs direct SDK:                                  

  - Provider has official Go SDK (OpenAI, Anthropic, Ollama) → Use direct SDK                                                         
  - Provider is OpenAI-compatible without Go SDK → Use openai.NewBase()

### CODE REVIEW

File: providers/openai/compatible.go
Comment (peteski22) [inline providers/openai/compatible.go:1]: ### Testing:

It looks like we're missing an entire test suite for this 😞 

## Organisation:

If this is `openai` compatibility stuff, it should be moved to the `openai` provider, with the `openai` provider becoming a thin wrapper.

e.g.

```go
func New(opts ...config.Option) (*Provider, error) {
    return NewBase(Defaults{
        Name:          "openai",
        DefaultURL:    "https://api.openai.com/v1",
        APIKeyEnvVar:  "OPENAI_API_KEY",
        RequireAPIKey: true,
        Capabilities:  capabilities,
    }, opts...)
}
```

(Move the base into `providers/openai/` and export it, and other providers that are openai compatible import it:

```go
import "github.com/mozilla-ai/any-llm-go/providers/openai"
```

So `llamafile` provider uses that.

providers/llamafile/llamafile.go
peteski22, line 1: Missing `var` interface assertions (see: `openai.go`): 

```go
// Ensure Provider implements the required interfaces.
var (
	_ providers.CapabilityProvider = (*Provider)(nil)
	_ providers.EmbeddingProvider  = (*Provider)(nil)
	_ providers.ErrorConverter     = (*Provider)(nil)
	_ providers.ModelLister        = (*Provider)(nil)
	_ providers.Provider           = (*Provider)(nil)
)
```

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+}
+
+// Ensure Provider implements the required interfaces.
+var (
+	_ providers.CapabilityProvider = (*Provider)(nil)
+	_ providers.EmbeddingProvider  = (*Provider)(nil)
+	_ providers.ErrorConverter     = (*Provider)(nil)
+	_ providers.ModelLister        = (*Provider)(nil)
+	_ providers.Provider           = (*Provider)(nil)
+)
```
peteski22: Shift this up above the type (and below the const) so we keep the layout in the files the same.

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+
+	// RequireAPIKey indicates whether an API key is required.
+	RequireAPIKey bool
+
+	// DefaultAPIKey is used when RequireAPIKey is false (e.g., for local servers).
+	DefaultAPIKey string
+
+	// Capabilities describes what the provider supports.
+	Capabilities providers.Capabilities
+}
```
peteski22: Order struct fields A-Z (don't need to optimize for padding).

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	if format == nil {
+		return openai.ChatCompletionNewParamsResponseFormatUnion{}
+	}
+
+	switch format.Type {
+	case "json_object":
+		return openai.ChatCompletionNewParamsResponseFormatUnion{
+			OfJSONObject: &openai.ResponseFormatJSONObjectParam{},
+		}
+	case "json_schema":
```
peteski22: Extract these to unexported package consts.

```go
const (
    responseFormatJSONObject = "json_object"
    responseFormatJSONSchema = "json_schema"
)
```

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+
+	// Check for OpenAI API error type.
+	var apiErr *openai.Error
+	if stderrors.As(err, &apiErr) {
+		return p.convertAPIError(apiErr, err)
+	}
+
+	// Network-level errors (connection refused, etc.).
+	if strings.Contains(err.Error(), "connection refused") {
```
peteski22: Seems a bit fragile checking strings like this, could we check for specific error types? If not then a comment which highlights the limitations should be added.

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	_ providers.CapabilityProvider = (*Provider)(nil)
+	_ providers.EmbeddingProvider  = (*Provider)(nil)
+	_ providers.ErrorConverter     = (*Provider)(nil)
+	_ providers.ModelLister        = (*Provider)(nil)
+	_ providers.Provider           = (*Provider)(nil)
+)
+
+// Provider implements the providers.Provider interface for OpenAI-compatible APIs.
+type Provider struct {
+	client      *openai.Client
```
peteski22: i don't think we need the pointer here, the SDK returns a value type anyway.

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	config      *config.Config
```
peteski22: Is this actually used other than being assigned in the constructor?

providers/openai/compatible.go
```diff
@@ -0,0 +1,624 @@
...
+
+	apiKey := cfg.ResolveAPIKey(providerCfg.APIKeyEnvVar)
+	if apiKey == "" && providerCfg.RequireAPIKey {
+		return nil, errors.NewMissingAPIKeyError(providerCfg.Name, providerCfg.APIKeyEnvVar)
+	}
+	if apiKey == "" {
+		apiKey = providerCfg.DefaultAPIKey
+	}
+
+	clientOpts := []option.RequestOption{
```
peteski22, line 116: I think there's something odd going on here with how the options are passed in but then handled.

e.g. timeout/HTTP client options seem to be ignored.. a caller can pass `WithTimeout()` or `WithHTTPClient()` but they have no effect. 🐛 Looks present in the `openai` provider from earlier PRs too 😞.

providers/openaicompat/openaicompat.go
peteski22: Create interface for provider defaults (in the config package).

```go
type ProviderDefaults interface {
    Capabilities() providers.Capabilities
    DefaultBaseURL() string
    Name() string
    ResolveAPIKey(cfg *Config) (string, error)
}

type Config struct {
    APIKey   string
    BaseURL  string
    Defaults ProviderDefaults  // interface <- nil if not set
    Timeout  time.Duration
}
```

Then each provider type (OpenAI-compatible, Anthropic-style, Ollama-style) implements the interface. This will scale to other standards without polluting config.

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+// New creates a new OpenAI-compatible provider.
+func New(providerCfg ProviderConfig, opts ...config.Option) (*Provider, error) {
+	cfg, err := config.New(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("invalid options: %w", err)
+	}
+
+	baseURL := cfg.BaseURL
+	if baseURL == "" && providerCfg.BaseURLEnvVar != "" {
+		baseURL = cfg.ResolveAPIKey(providerCfg.BaseURLEnvVar)
```
peteski22: `ResolveAPIKey()` for a base URL? 🤔

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+
+// New creates a new OpenAI-compatible provider.
+func New(providerCfg ProviderConfig, opts ...config.Option) (*Provider, error) {
+	cfg, err := config.New(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("invalid options: %w", err)
+	}
+
+	baseURL := cfg.BaseURL
+	if baseURL == "" && providerCfg.BaseURLEnvVar != "" {
```
peteski22: Probably should do some validation on these things like:
* Trimming the white space before `!= ""`
* URL validation on URLs
* If the env var is set but empty what happens?

Something like:

```go
func (c *Config) ResolveEnv(envVar string) string  // generic env resolver
func (c *Config) ResolveBaseURL(envVar, defaultVal string) (string, error)  // with validation
```

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	go func() {
+		defer close(chunks)
+		defer close(errs)
+
+		req := convertParams(params)
+		stream := p.client.Chat.Completions.NewStreaming(ctx, req)
+
+		for stream.Next() {
+			chunk := stream.Current()
+			chunks <- convertChunk(&chunk)
```
peteski22: Missed this on the openai one earlier too:

If the consumer abandons the channel (e.g., context cancelled, error elsewhere), this goroutine blocks forever on the send. Should use select with context:

```go
    select {
    case chunks <- convertChunk(&chunk):
    case <-ctx.Done():
        return
}
```

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	case providers.RoleSystem:
+		return openai.SystemMessage(msg.ContentString())
+	case providers.RoleUser:
+		return convertUserMessage(msg)
+	case providers.RoleAssistant:
+		return convertAssistantMessage(msg)
+	case providers.RoleTool:
+		return openai.ToolMessage(msg.ToolCallID, msg.ContentString())
+	default:
+		return openai.UserMessage(msg.ContentString())
```
peteski22: Missed this too in earlier reviews 😞 

Unknown roles silently become user messages. This hides bugs. Should either error or log a warning.

providers/openai/compatible.go
```diff
@@ -0,0 +1,624 @@
...
+func convertMessages(messages []providers.Message) []openai.ChatCompletionMessageParamUnion {
+	result := make([]openai.ChatCompletionMessageParamUnion, 0, len(messages))
+	for _, msg := range messages {
+		result = append(result, convertMessage(msg))
+	}
+	return result
+}
+
+// convertParams converts providers.CompletionParams to OpenAI request parameters.
+func convertParams(params providers.CompletionParams) openai.ChatCompletionNewParams {
```
peteski22, line 459: Should have input validation, should validate Model is non-empty and Messages has at least one entry.

providers/openai/compatible.go
```diff
@@ -0,0 +1,624 @@
...
+	return result
+}
+
+// convertEmbeddingParams converts provider embedding params to OpenAI format.
+func convertEmbeddingParams(params providers.EmbeddingParams) openai.EmbeddingNewParams {
+	req := openai.EmbeddingNewParams{
+		Model: openai.EmbeddingModel(params.Model),
+	}
+
+	switch v := params.Input.(type) {
```
peteski22, line 369: Should probably have a `default` case... Should error or have an explicit default.

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+	var apiErr *openai.Error
+	if stderrors.As(err, &apiErr) {
+		return p.convertAPIError(apiErr, err)
+	}
+
+	// Network-level errors (connection refused, etc.).
+	if strings.Contains(err.Error(), "connection refused") {
+		return errors.NewProviderError(
+			p.providerCfg.Name,
+			fmt.Errorf("%s server not running: %w", p.providerCfg.Name, err),
```
peteski22: Doesn't need the provider name in the error msg as the `Error()` interface implementation includes it in the final error:

  ```go
  func (e *BaseError) Error() string {
	msg := ""
	if e.Err != nil {
		msg = e.Err.Error()
	}
	if e.Provider != "" {
		return fmt.Sprintf("[%s] %s: %s", e.Provider, e.Code, msg)
	}
	return fmt.Sprintf("%s: %s", e.Code, msg)
}
```

File: providers/llamafile/llamafile_test.go [hunk 15]
```diff
@@ -0,0 +1,232 @@
...
+	params := providers.CompletionParams{
+		Model:    testutil.TestModel(providerName),
+		Messages: testutil.SimpleMessages(),
+	}
+
+	resp, err := provider.Completion(ctx, params)
+	require.NoError(t, err)
+
+	require.NotEmpty(t, resp.ID)
+	require.Equal(t, "chat.completion", resp.Object)
```
peteski22: Missed this on other PRs too, but should use the consts vs. strings for consistency and to make refactoring  (e.g. `objectChatCompletion`),

providers/openaicompat/openaicompat.go
```diff
@@ -0,0 +1,624 @@
...
+			},
+		})
+	}
+	return result
+}
+
+// convertUserMessage converts a user message to OpenAI format.
+func convertUserMessage(msg providers.Message) openai.ChatCompletionMessageParamUnion {
+	if msg.IsMultiModal() {
+		parts := make([]openai.ChatCompletionContentPartUnionParam, 0)
```
peteski22: Could use `make(..., 0, len(msg.ContentParts()))` for efficiency like elsewhere in the code.

peteski22 (changes requested):
Added comments, also a general observation that we have no logging anywhere in the package. We should consider adding optional structured logging (at least for errors and unusual conditions).

providers/openai/base.go
```diff
@@ -0,0 +1,711 @@
...
+	_ providers.CapabilityProvider = (*BaseProvider)(nil)
+	_ providers.EmbeddingProvider  = (*BaseProvider)(nil)
+	_ providers.ErrorConverter     = (*BaseProvider)(nil)
+	_ providers.ModelLister        = (*BaseProvider)(nil)
+	_ providers.Provider           = (*BaseProvider)(nil)
+)
+
+// BaseProvider implements the providers.Provider interface for OpenAI-compatible APIs.
+// It can be embedded by other providers that use OpenAI-compatible endpoints.
+type BaseProvider struct {
```
peteski22:
```suggestion
type CompatibleProvider struct {
```

providers/openai/base.go
```diff
@@ -0,0 +1,711 @@
...
+
+// Response format types.
+const (
+	responseFormatJSONObject = "json_object"
+	responseFormatJSONSchema = "json_schema"
+)
+
+// BaseConfig contains the configuration for an OpenAI-compatible provider.
+// Fields are ordered alphabetically.
+type BaseConfig struct {
```
peteski22:
```suggestion
type CompatibleConfig struct {
```

providers/openai/base.go
```diff
@@ -0,0 +1,711 @@
...
+
+// BaseProvider implements the providers.Provider interface for OpenAI-compatible APIs.
+// It can be embedded by other providers that use OpenAI-compatible endpoints.
+type BaseProvider struct {
+	baseConfig BaseConfig
+	client     openai.Client
+}
+
+// NewBase creates a new OpenAI-compatible base provider.
+func NewBase(baseCfg BaseConfig, opts ...config.Option) (*BaseProvider, error) {
```
peteski22:
```suggestion
func NewCompatible(baseCfg CompatibleConfig, opts ...config.Option) (*CompatibleProvider, error) {
```

providers/openai/compatible.go
peteski22, line 1: Nit: maybe rename to `compatible.go` (and `compatible_test.go`)

providers/openai/base.go
```diff
@@ -0,0 +1,711 @@
...
+	return cfg.APIKey
+}
+
+// resolveBaseURL resolves the base URL from config or environment.
+func resolveBaseURL(cfg *config.Config, baseCfg BaseConfig) string {
+	if cfg.BaseURL != "" {
+		return cfg.BaseURL
+	}
+	if baseCfg.BaseURLEnvVar != "" {
+		if url := cfg.ResolveAPIKey(baseCfg.BaseURLEnvVar); url != "" {
```
peteski22: This still needs to be looked at (per my previous review), if there's some reason we should be doing this, please add a comment to explain.

providers/openai/base.go
```diff
@@ -0,0 +1,711 @@
...
+func resolveAPIKey(cfg *config.Config, baseCfg BaseConfig) string {
+	if baseCfg.APIKeyEnvVar != "" {
+		return cfg.ResolveAPIKey(baseCfg.APIKeyEnvVar)
+	}
+	return cfg.APIKey
+}
+
+// resolveBaseURL resolves the base URL from config or environment.
+func resolveBaseURL(cfg *config.Config, baseCfg BaseConfig) string {
+	if cfg.BaseURL != "" {
```
peteski22: Previous review mentioned validating params more carefully (trim whitespace, check valid URL formats etc) - still stands.

peteski22 (changes requested):
Just added a few last bits.. looking good, almost there.

peteski22 (approved):
🦙

PR #24, opened 2026-02-02
feat(provider): add Google Gemini provider

agpituk:
## Summary

Add Google Gemini provider using the official `google.golang.org/genai` SDK. This enables completion, streaming, image understanding, reasoning/thinking, embeddings, and model listing through the unified any-llm interface.

## Changes

- Add `providers/gemini/gemini.go` - Core provider implementation
- Add `providers/gemini/gemini_test.go` - Comprehensive unit and integration tests
- Update `internal/testutil/fixtures.go` - Update Gemini model mappings to match Python any-llm
- Update `docs/providers.md` - Document Gemini provider with usage examples
- Update `go.mod` / `go.sum` - Add genai SDK dependency

## Interfaces Implemented

- `Provider`
- `CapabilityProvider`
- `EmbeddingProvider`
- `ModelLister`
- `ErrorConverter`

## Testing

- 20 unit tests covering message conversion, finish reasons, tools, tool choice, error mapping, thinking budgets, streaming state, response conversion, image parts, embedding input, ID generation, response format
- 6 integration tests (run when `GEMINI_API_KEY` is set)
- All tests pass with `gemini-2.5-flash` and `gemini-embedding-001` models

```
$ go test ./providers/gemini/ -count=1 -short
ok  	github.com/mozilla-ai/any-llm-go/providers/gemini	0.919s
```

```
$ go test ./... -count=1 -short
ok  	github.com/mozilla-ai/any-llm-go/config	0.646s
ok  	github.com/mozilla-ai/any-llm-go/errors	0.364s
ok  	github.com/mozilla-ai/any-llm-go/providers/anthropic	1.088s
ok  	github.com/mozilla-ai/any-llm-go/providers/gemini	0.919s
ok  	github.com/mozilla-ai/any-llm-go/providers/llamafile	2.943s
ok  	github.com/mozilla-ai/any-llm-go/providers/ollama	4.051s
ok  	github.com/mozilla-ai/any-llm-go/providers/openai	2.297s
ok  	github.com/mozilla-ai/any-llm-go/providers/platform	2.378s
```

## Checklist

- [x] Documentation updated
- [x] No breaking changes


peteski22 (commented):
Good implementation overall. Left some inline comments on a few areas that could use hardening around error handling.

providers/gemini/gemini.go
```diff
@@ -0,0 +1,790 @@
...
+		}
+	}
+
+	return genai.NewContentFromParts(parts, roleUser)
+}
+
+// generateID generates a random ID with the given prefix.
+func generateID(prefix string) string {
+	b := make([]byte, 12)
+	_, _ = rand.Read(b)
```
peteski22: The error from `crypto/rand.Read` is silently discarded here. While failures are rare, entropy exhaustion could lead to predictable or zero-filled IDs.

The idiomatic approach would be to propagate the error up the call chain:

```go
func generateID(prefix string) (string, error) {
    b := make([]byte, 12)
    if _, err := rand.Read(b); err != nil {
        return "", fmt.Errorf("generating ID: %w", err)
    }
    return prefix + hex.EncodeToString(b), nil
}
```

This would require updating callers to handle the error, which is a larger change. Alternatively, if the current signature needs to be preserved, you could log a warning and fall back to a timestamp-based ID - though that's less ideal from a uniqueness perspective.

providers/gemini/gemini.go
```diff
@@ -0,0 +1,790 @@
...
+				return genai.NewPartFromBytes(data, mediaType)
+			}
+		}
+	}
+
+	return &genai.Part{
+		FileData: &genai.FileData{
+			FileURI:  url,
+			MIMEType: "image/jpeg",
+		},
```
peteski22, line 606: When base64 decoding fails for what appears to be a data URL, the function silently falls through to treating it as a FileURI. This could mask data URL parsing errors and lead to confusing downstream failures.

Additionally, the `"image/jpeg"` MIME type is a magic string that should be extracted to a constant (e.g., `defaultImageMIMEType`).

Consider returning an error if base64 decoding fails for a data URL, or at minimum add a comment explaining this fallback behavior is intentional.

providers/gemini/gemini.go
```diff
@@ -0,0 +1,790 @@
...
+	switch msg.Role {
+	case providers.RoleUser:
+		return convertUserMessage(msg)
+	case providers.RoleAssistant:
+		return convertAssistantMessage(msg)
+	case providers.RoleTool:
+		return convertToolMessage(msg)
+	default:
+		return nil
+	}
```
peteski22, line 623: Returning `nil` for unknown roles could cause messages to be silently dropped from conversations. Per project guidelines, unknown enum values should not be silently converted.

If this is intentional behavior (e.g., to gracefully handle future role types), consider adding a comment explaining why, or log a warning when an unknown role is encountered.

providers/gemini/gemini.go
```diff
@@ -0,0 +1,790 @@
...
+
+	switch apiErr.Code {
+	case 401, 403:
+		return errors.NewAuthenticationError(providerName, err)
+	case 404:
+		return errors.NewModelNotFoundError(providerName, err)
+	case 429:
+		return errors.NewRateLimitError(providerName, err)
+	case 400:
+		msg := strings.ToLower(apiErr.Message)
```
peteski22, line 252: The 400 error handling uses string matching on error messages, which is fragile and could break if Gemini changes their error message formats.

If the Gemini SDK doesn't expose typed errors for these specific conditions (context length, content filter), this may be a pragmatic necessity. In that case, consider:
1. Adding a comment documenting this as a known limitation
2. Extracting the matched strings to constants (e.g., `errMsgContext`, `errMsgSafety`)

This would make future SDK updates easier to accommodate.

peteski22 (approved):

PR #25, opened 2026-02-03
fix(anthropic): handle []string type for tool required fields & improving overal tests

agpituk:
## Summary

- Fix bug where tool `required` fields defined as `[]string` in Go code were silently ignored
- Add comprehensive test coverage for tool conversion and agent loop scenarios

## Changes

- **fix(anthropic)**: Add type switch in `convertTool` to handle both `[]string` (Go-defined) and `[]any` (JSON-unmarshaled) for the required field
- **test(fixtures)**: Add `CalculatorTool`, `MockWeatherResult`, and `MockCalculatorResult` helpers
- **test(anthropic)**: Add `TestConvertTool` unit tests and `TestIntegrationAgentLoop` / `TestIntegrationAgentLoopMultipleParams` integration tests
- **test(openai)**: Add `TestConvertTools` unit tests and matching agent loop integration tests

## Testing

- `make test-unit` - all unit tests pass
- `make test` - integration tests verify full agent loop with real API calls

## Checklist

### DIFF
--- a/internal/testutil/fixtures.go
+++ b/internal/testutil/fixtures.go
@@ -2,6 +2,7 @@
 package testutil
 
 import (
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/testutil/fixtures.go
```diff
@@ -179,6 +180,69 @@ func DateTool() providers.Tool {
 	}
 }
 
+// CalculatorTool returns a calculator tool with multiple parameters for testing.
+// This tool is useful for verifying that parameter order and required fields
+// are correctly preserved during conversion.
+func CalculatorTool() providers.Tool {
+	return providers.Tool{
```
peteski22: In test code we should always make sure that we're really clear with the naming, so `TestSomething` or `MockSomething` etc.

If this is a helper function then we should also make sure we're:

* Supplying `t *testing.T` as the 1st parameter to the function
* Calling `t.Helper()` as the first thing in the function


```suggestion
func NewTestCalculatorTool(t *testing.T) providers.Tool {
    t.Helper()

	return providers.Tool{
```

https://cs.opensource.google/go/go/+/refs/tags/go1.25.6:src/testing/testing.go;l=1260-1262

Will need to update Go doc and call sites.

internal/testutil/fixtures.go
```diff
@@ -179,6 +180,69 @@ func DateTool() providers.Tool {
...
+	return `{"result": ` + formatFloat(result) + `}`
+}
+
+// formatFloat formats a float for JSON output.
+func formatFloat(f float64) string {
+	if f == float64(int(f)) {
+		return fmt.Sprintf("%.0f", f)
+	}
+	return fmt.Sprintf("%g", f)
+}
```
peteski22: Pass in `t *testing.T` for `t.Helper()`.

providers/anthropic/anthropic.go
```diff
@@ -524,11 +524,15 @@ func convertTool(tool providers.Tool) anthropic.ToolUnionParam {
 		inputSchema.Properties = props
 	}
 	if req, ok := tool.Function.Parameters["required"]; ok {
```
peteski22: You should check if `tool.Function.Parameters` is `nil` (as it's a map, which is represented by a pointer).

PR #26, opened 2026-02-04
feat(provider): add DeepSeek provider

agpituk:
## Summary

Add DeepSeek provider implementation using the OpenAI-compatible base with custom handling for JSON schema responses.

## Changes

- Add `providers/deepseek/deepseek.go` - Provider implementation
- Add `providers/deepseek/deepseek_test.go` - Comprehensive test suite

### Features

- Chat completion and streaming
- Tool calling support
- JSON schema support (via preprocessing)
- List models
- Reasoning support (DeepSeek R1)

### JSON Schema Handling

DeepSeek doesn't support `json_schema` response format directly. Instead, it requires `json_object` mode with the schema embedded in the user message. This implementation automatically converts `json_schema` requests by:

1. Injecting the schema into the last user message
2. Converting `response_format` to `json_object`
3. Preserving original params if injection fails (no user message, multimodal content, or marshal error)

See: https://api-docs.deepseek.com/guides/json_mode

### Unit Tests
- Provider creation and API key validation
- Capabilities reporting
- JSON schema preprocessing (success and failure cases)
- Message field preservation (Reasoning, multimodal detection)

### Integration Tests
- Basic completion and streaming
- System messages and conversations
- JSON schema responses
- Tool calling
- Agent loops with weather and calculator tools

```bash
# Run unit tests
go test -v ./providers/deepseek/... -run 'Test[^I]'

# Run integration tests (requires DEEPSEEK_API_KEY)
go test -v ./providers/deepseek/... -run 'TestIntegration'
```

All tests passing:
```
ok      github.com/mozilla-ai/any-llm-go/providers/deepseek    3.209s
```

## Checklist

### DIFF
--- a/providers/deepseek/deepseek.go
+++ b/providers/deepseek/deepseek.go
@@ -20,6 +20,13 @@ const (
 	providerName   = "deepseek"
 )
 
+// Object type constants for API responses.
+const (
+	objectChatCompletion      = "chat.completion"
+	objectChatCompletionChunk = "chat.completion.chunk"
+	objectList                = "list"
+)
+
 // Response format types.
 const (
 	responseFormatJSONObject = "json_object"
--- a/providers/deepseek/deepseek.go
+++ b/providers/deepseek/deepseek.go
@@ -1,0 +1,210 @@
+// Package deepseek provides a DeepSeek provider implementation for any-llm.
+// DeepSeek exposes an OpenAI-compatible API with some differences in JSON mode handling.
+package deepseek
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/mozilla-ai/any-llm-go/config"
+	"github.com/mozilla-ai/any-llm-go/providers"
+	"github.com/mozilla-ai/any-llm-go/providers/openai"
+)
+
+// Provider configuration constants.
+const (
+	defaultBaseURL = "https://api.deepseek.com"
+	envAPIKey      = "DEEPSEEK_API_KEY"
+	providerName   = "deepseek"
+)
+
+// Response format types.
+const (
+	responseFormatJSONObject = "json_object"
+	responseFormatJSONSchema = "json_schema"
+)
+
+// Ensure Provider implements the required interfaces.
+var (
+	_ providers.CapabilityProvider = (*Provider)(nil)
+	_ providers.ErrorConverter     = (*Provider)(nil)
+	_ providers.ModelLister        = (*Provider)(nil)
+	_ providers.Provider           = (*Provider)(nil)
+)
+
+// Provider implements the providers.Provider interface for DeepSeek.
+// It embeds openai.CompatibleProvider since DeepSeek exposes an OpenAI-compatible API.
+type Provider struct {
+	*openai.CompatibleProvider
+}
+
+// New creates a new DeepSeek provider.
+func New(opts ...config.Option) (*Provider, error) {
+	base, err := openai.NewCompatible(openai.CompatibleConfig{
+		APIKeyEnvVar:   envAPIKey,
+		BaseURLEnvVar:  "",
+		Capabilities:   deepseekCapabilities(),
+		DefaultAPIKey:  "",
+		DefaultBaseURL: defaultBaseURL,
+		Name:           providerName,
+		RequireAPIKey:  true,
+	}, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Provider{CompatibleProvider: base}, nil
+}
+
+// Completion performs a chat completion request.
+// It overrides the base implementation to handle DeepSeek's JSON mode quirks.
+func (p *Provider) Completion(
+	ctx context.Context,
+	params providers.CompletionParams,
+) (*providers.ChatCompletion, error) {
+	params = preprocessParams(params)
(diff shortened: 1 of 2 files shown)


providers/deepseek/deepseek.go
```diff
@@ -0,0 +1,210 @@
...
+	// Create the modified content with JSON instructions.
+	modifiedContent := fmt.Sprintf(`Please respond with a JSON object that matches the following schema:
+
+%s
+
+Return the JSON object only, no other text, do not wrap it in `+"```json"+` or `+"```"+`.
+
+%s`, string(schemaJSON), originalContent)
+
+	// Create a copy of messages to avoid mutating the original.
```
peteski22, line 203: You can use `slices.Clone` here instead of creating the `result` slice and `copy`. 

Note: both `slices.Clone` and `copy` do a shallow copy, is that what you want (vs. deep copy)?

providers/deepseek/deepseek_test.go
```diff
@@ -0,0 +1,713 @@
...
+	"github.com/mozilla-ai/any-llm-go/internal/testutil"
+	"github.com/mozilla-ai/any-llm-go/providers"
+)
+
+// Expected object types in API responses.
+const (
+	objectChatCompletion      = "chat.completion"
+	objectChatCompletionChunk = "chat.completion.chunk"
+	objectList                = "list"
+)
```
peteski22: If these are valid for prod code, move them to to the `deepseek.go` file as consts, hiding non-test-named things in test files isn't good (they're in the same package so accessible inside `deepseek.go`.

providers/deepseek/deepseek_test.go
```diff
@@ -0,0 +1,713 @@
...
+		require.Equal(t, "thinking...", result[0].Reasoning.Content)
+	})
+}
+
+// Integration tests - only run if DeepSeek API key is available.
+
+func TestIntegrationCompletion(t *testing.T) {
+	t.Parallel()
+
+	if testutil.SkipIfNoAPIKey(providerName) {
```
peteski22, line 361: Not for this PR, but the fixtures stuff should be reviewed, as this could probably just be a test helper that handles the skip itself, vs. having the:

```go
if testutil.SkipIfNoAPIKey(providerName) {
    t.Skip("FOO_API_KEY not set")
}
```

everywhere

providers/deepseek/deepseek_test.go
```diff
@@ -0,0 +1,713 @@
...
+
+	t.Run("does not mutate original messages", func(t *testing.T) {
+		t.Parallel()
+
+		messages := []providers.Message{
+			{Role: providers.RoleUser, Content: "Original content"},
+		}
+		schema := map[string]any{"type": "object"}
+
+		_, _ = preprocessMessagesForJSONSchema(messages, schema)
```
peteski22, line 330: Add a comment to explain why we just ditch the values in this test please.

providers/deepseek/deepseek_test.go
```diff
@@ -0,0 +1,713 @@
...
+	resp, err := provider.Completion(ctx, params)
+	require.NoError(t, err)
+
+	require.NotEmpty(t, resp.ID)
+	require.Len(t, resp.Choices, 1)
+
+	// Should either have tool calls or content.
+	choice := resp.Choices[0]
+	hasToolCalls := len(choice.Message.ToolCalls) > 0
+	hasContent := choice.Message.Content != nil && choice.Message.Content != ""
```
peteski22:
```suggestion
	hasContent := choice.Message.ContentString() != ""
```

peteski22 (approved):
Some nit picking comments, but looks good. 

I'd be super happy if we could address them, but I'm not going to block this on it. 😄

providers/deepseek/deepseek_test.go
peteski22: https://github.com/mozilla-ai/any-llm-go/issues/28

providers/deepseek/deepseek_test.go
agpituk: These constants (`objectChatCompletion`, `objectChatCompletionChunk`, `objectList`) are intentionally test-only - they're used to assert the expected response types from the API, but aren't needed in production code (we don't branch on these values, just pass them through).

This matches the pattern in other providers like `anthropic` and `openai` where similar assertion constants live in test files.

That said, if you feel strongly about it I'm happy to move them - just let me know!

providers/deepseek/deepseek_test.go
agpituk: Agreed - having the helper handle the skip internally would reduce boilerplate. I see you've already tracked this in #28, so I'll leave it for a follow-up.

agpituk:
> Some nit picking comments, but looks good.
> 
> I'd be super happy if we could address them, but I'm not going to block this on it. 😄

thanks Peter, I fixed the majority of them, re-requested review in case you're not happy with them, if not, we can merge :)

providers/deepseek/deepseek_test.go
peteski22: I'm not sure what you meant here - the openai provider has these constants in `compatible.go` (production code), and anthropic uses string literals directly without constants. Neither has them defined in test files.

Either way, I've moved them to the production file now to match the openai pattern, so we're all good. 👍

PR #42, opened 2026-02-11
feat(providers): add CompletionTools field to Capabilities struct

### LINKED ISSUE #39: Add CompletionTools field to Capabilities struct
## Problem

The `Capabilities` struct in `providers/types.go` has explicit fields for completion modifiers (`CompletionImage`, `CompletionPDF`, `CompletionReasoning`, `CompletionStreaming`) but no field for tool/function calling support.

Tool support is currently tracked only in the manually-maintained tables in `README.md` and `docs/providers.md`. There's no way for callers to programmatically check whether a provider supports tool calling.

## Proposal

Add a `CompletionTools bool` field to the `Capabilities` struct:

```go
type Capabilities struct {
    Completion          bool
    CompletionImage     bool
    CompletionPDF       bool
    CompletionReasoning bool
    CompletionStreaming  bool
    CompletionTools     bool
    Embedding           bool
    ListModels          bool
}
```

Then set it appropriately in each provider's capabilities function.

peteski22:
## Summary

- Adds `CompletionTools bool` field to the `Capabilities` struct so callers can programmatically check tool/function calling support
- Sets `CompletionTools: true` for all providers (matching the existing docs tables)
- Updates all provider tests with capability assertions

Closes #39

## Test plan

- [x] All unit tests pass (`make test-unit`)
- [x] Linting passes (`golangci-lint run --fix -v`)
- [x] Each commit is atomic: one per provider with its test

### DIFF
--- a/internal/testutil/mocks.go
+++ b/internal/testutil/mocks.go
@@ -122,6 +122,7 @@ func NewMockProvider() *MockProvider {
 			return providers.Capabilities{
 				Completion:          true,
 				CompletionStreaming: true,
+				CompletionTools:     true,
 				Embedding:           true,
 				ListModels:          true,
 			}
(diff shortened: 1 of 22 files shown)


tbille (approved):
🚢 🇮🇹

PR #60, opened 2026-03-01
feat(gemini): ThoughtSignature round-trip + fix 5xx transient error retry

mrbeandev:
## Summary

Two fixes for the Gemini provider, both discovered while using any-llm-go in a long-running agentic loop with Gemini 2.5 Pro thinking models.

### 1. ThoughtSignature round-trip support

**Problem:** Gemini 2.5+ thinking models attach a `ThoughtSignature` to every `FunctionCall` part in the response. When conversation history is replayed on the next turn, the API requires these signatures to be echoed back verbatim. Without this, the API returns a `400 Bad Request` error on the second tool-call turn and beyond.

**Fix:**
- `providers/types.go`: Added a `ThoughtSignature []byte` field (tagged `json:"-"`) to `ToolCall`. The field is excluded from JSON marshalling so it does not affect the wire format; callers round-trip it through whatever storage mechanism they use.
- `providers/gemini/gemini.go`:
  - **Capture (streaming):** `streamState.processResponse` now copies `part.ThoughtSignature → toolCall.ThoughtSignature` for each function-call part.
  - **Capture (non-streaming):** `extractResponseContent` does the same.
  - **Replay:** `convertAssistantMessage` sets `part.ThoughtSignature` from `ToolCall.ThoughtSignature` when echoing history. Falls back to the `geminiThoughtSignatureBypass` sentinel constant (documented by Google) for older tool calls recorded without a signature, to preserve backward compatibility.

### 2. Transient 5xx error retry

**Problem:** HTTP 500, 502, 503, and 529 responses from the Gemini API fell through `ConvertError` without a matching case and became `ProviderError`. Since callers typically only retry on `RateLimitError`, these transient server-side errors were not retried, causing unnecessary hard failures under load.

**Fix:** Added `case 500, 502, 503, 529` in `ConvertError` mapping to `NewRateLimitError`, consistent with how callers already handle 429.

## Testing

- Manually tested with live Gemini 2.5 Pro (thinking enabled) in a multi-turn agentic loop — no more 400 errors on tool-call replay.
- Manually reproduced and verified the 503 retry fix against a Gemini endpoint under load.
- The changes are also running in production in [only-true-agi](https://github.com/mrbeandev/only-true-agi), an agentic framework built on top of this library.

## Breaking changes

None. `ThoughtSignature` is `json:"-"` so existing serialized `ToolCall` values remain valid. Error classification for 5xx is strictly more permissive (adds retry, no new failures).

### DIFF
--- a/providers/gemini/gemini.go
+++ b/providers/gemini/gemini.go
@@ -73,6 +73,16 @@ const (
 // Default MIME type for image URLs when type cannot be determined.
 const defaultImageMIMEType = "image/jpeg"
 
+// geminiThoughtSignatureBypass is a sentinel value that can be used as a
+// ThoughtSignature when no real signature is available. Gemini 2.5+ thinking
+// models require a ThoughtSignature on every tool-call part when replaying
+// conversation history. If a call was recorded without a signature (e.g. from
+// an earlier session or a non-thinking model), supplying this bypass value
+// tells the API to skip signature validation for that part.
+//
+// Reference: https://ai.google.dev/gemini-api/docs/thinking#thought-signatures
+const geminiThoughtSignatureBypass = "context_engineering_is_the_way_to_go"
+
 // Error message patterns for 400 error classification.
 // The Gemini SDK doesn't expose typed errors for these conditions,
 // so we rely on message matching as a pragmatic fallback.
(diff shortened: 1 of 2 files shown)


peteski22:
Hi @mrbeandev — thank you for this PR! You identified a real and important issue with ThoughtSignature round-tripping on Gemini 2.5+ thinking models.

We've taken your work and built on it in #61, which:

- Uses a provider-namespaced `Extra` field on `ToolCall` (matching the existing `Extra` pattern on `CompletionParams`) rather than a dedicated `ThoughtSignature` field, to keep the shared types provider-agnostic
- Captures signatures in both streaming and non-streaming paths
- Adds the bypass fallback for tool calls without a real signature
- Includes comprehensive test coverage (15 test cases)
- Files an upstream issue on the Go SDK for the `[]byte` base64 encoding behaviour: https://github.com/googleapis/go-genai/issues/711

We also investigated the 5xx → `RateLimitError` mapping and opted to keep 5xx as `ProviderError` to match the Python sister library's semantics — server errors and rate limits are distinct failure modes.

You're credited as co-author on all commits. Closing this in favour of #61 — thanks again for the contribution!

PR #65, opened 2026-03-11
feat(config): add WithHeader option for custom HTTP headers

jrschumacher:
## Summary

- Add a `WithHeader` config option that injects custom HTTP headers on every provider request.
- Inject headers via a transport wrapper (`headerTransport`) that **clones each request before adding headers**, so the caller's `*http.Request` is never mutated (per the `net/http` `RoundTripper` contract).
- Route the Anthropic provider through `cfg.HTTPClient()` so custom headers (and the configured client) apply to Anthropic too.
- Re-export `WithHeader` from the root `anyllm` package.

`Headers` is stored as the stdlib `http.Header`, and `WithHeader` uses `Set` semantics — repeated calls with the same key overwrite the previous value. The header set is cloned when the client is built, so mutating `cfg.Headers` afterward cannot race an in-flight request.

## Motivation

Proxy/gateway services like [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/) require custom authentication headers (e.g., `cf-aig-authorization`) alongside the provider's own API key. Previously there was no way to inject custom headers without supplying a fully custom HTTP client with a hand-rolled round-tripper.

## ⚠️ Behavior change (Anthropic)

Before this PR the Anthropic provider used the `anthropic-sdk-go` default HTTP client, which has **no timeout**. It now uses `cfg.HTTPClient()`, so the configured `cfg.Timeout` (default **120s**) applies to Anthropic requests. Because this is an `http.Client.Timeout`, it bounds the entire request — **including streaming responses**. Callers whose Anthropic calls can exceed 120s should raise it with `WithTimeout`, or supply their own client via `WithHTTPClient`. This aligns Anthropic with every other provider, which already route through `cfg.HTTPClient()`.

## Usage

```go
provider, err := openai.New(
    anyllm.WithAPIKey("sk-..."),
    anyllm.WithBaseURL("https://gateway.ai.cloudflare.com/v1/{account}/{gateway}/openai"),
    anyllm.WithHeader("cf-aig-authorization", "Bearer cf-token"),
)
```

Works with all providers, since they all use `cfg.HTTPClient()` (Anthropic included, via this PR).

## Changes

| File | Change |
|------|--------|
| `config/config.go` | `Headers http.Header` field; `WithHeader` option (Set semantics, trims key); `headerTransport` round-tripper that clones the request before injecting headers; `HTTPClient()` wraps the base transport when headers are set (cloning the header set) while preserving the base client's `Timeout`/`CheckRedirect`/`Jar` |
| `config/config_test.go` | Tests: key/value validation, multiple headers, duplicate-key overwrite, transport wrapping (default + custom client), no-op when no headers set, and an end-to-end `httptest` test asserting inbound headers and that the caller's request is not mutated |
| `providers/anthropic/anthropic.go` | Pass `cfg.HTTPClient()` to the SDK client options |
| `anyllm.go` | Re-export `WithHeader` |

## Test plan

- [x] `go build ./...`
- [x] `go test -race ./config/...` (unit + end-to-end)
- [x] `golangci-lint run` clean
- [x] Rebased onto latest `main`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,6 +18,11 @@ type Config struct {
 	// BaseURL is the base URL for the API. If empty, the provider's default is used.
 	BaseURL string
 
+	// Headers holds extra HTTP headers to include on every request.
+	// Useful for proxy authentication (e.g., Cloudflare AI Gateway's
+	// cf-aig-authorization header).
+	Headers map[string]string
+
 	// Extra holds provider-specific configuration options.
 	Extra map[string]any
 
@@ -90,6 +95,26 @@ func WithBaseURL(baseURL string) Option {
 	}
 }
 
+
+// WithHeader adds an HTTP header that will be included on every request.
+// Useful for proxy/gateway authentication (e.g., cf-aig-authorization for
+// Cloudflare AI Gateway). Can be called multiple times to add multiple headers.
+func WithHeader(key, value string) Option {
+	return func(c *Config) error {
+		key = strings.TrimSpace(key)
+		if key == "" {
+			return fmt.Errorf("header key cannot be empty")
+		}
+
+		if c.Headers == nil {
+			c.Headers = make(map[string]string)
+		}
+
+		c.Headers[key] = value
+		return nil
+	}
+}
+
 // WithExtra sets extra provider-specific configuration.
 // Whitespace is automatically trimmed from the key.
 func WithExtra(key string, value any) Option {
@@ -155,9 +180,41 @@ func (c *Config) HTTPClient() *http.Client {
 		if c.httpClient == nil {
 			c.httpClient = &http.Client{Timeout: c.Timeout}
 		}
+
+		// Wrap the transport to inject custom headers on every request.
+		if len(c.Headers) > 0 {
+			c.httpClient = &http.Client{
+				Transport: &headerTransport{
+					base:    c.httpClient.Transport,
+					headers: c.Headers,
+				},
+				Timeout:       c.httpClient.Timeout,
+				CheckRedirect: c.httpClient.CheckRedirect,
+				Jar:           c.httpClient.Jar,
+			}
+		}
 	})
 
 	return c.httpClient
+}
+
+// headerTransport is an http.RoundTripper that injects headers on every request.
+type headerTransport struct {
+	base    http.RoundTripper
+	headers map[string]string
+}
+
+func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	for key, value := range t.headers {
+		req.Header.Set(key, value)
+	}
+
+	base := t.base
+	if base == nil {
+		base = http.DefaultTransport
+	}
+
+	return base.RoundTrip(req)
 }
 
 // ResolveAPIKey returns the API key from config if set, otherwise falls back
--- a/config/config.go
+++ b/config/config.go
@@ -21,7 +21,7 @@ type Config struct {
 	// Headers holds extra HTTP headers to include on every request.
 	// Useful for proxy authentication (e.g., Cloudflare AI Gateway's
 	// cf-aig-authorization header).
-	Headers map[string]string
+	Headers http.Header
 
 	// Extra holds provider-specific configuration options.
 	Extra map[string]any
@@ -37,6 +37,33 @@ type Config struct {
 
 // Option is a function that modifies the Config.
 type Option func(*Config) error
+
+// headerTransport is an http.RoundTripper that injects a fixed set of headers
+// onto every request it forwards.
+type headerTransport struct {
+	base    http.RoundTripper
+	headers http.Header
+}
+
+// RoundTrip injects the configured headers and forwards the request.
+// It operates on a clone so the caller's *http.Request is left unmodified,
+// per the http.RoundTripper contract.
+func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	clone := req.Clone(req.Context())
+	for key, values := range t.headers {
+		clone.Header.Del(key)
+		for _, value := range values {
+			clone.Header.Add(key, value)
+		}
+	}
+
+	base := t.base
+	if base == nil {
+		base = http.DefaultTransport
+	}
+
+	return base.RoundTrip(clone)
+}
 
 // New creates a Config with the given options applied.
 // Note: HTTPClient is not created here by default; it is lazily created via the HTTPClient()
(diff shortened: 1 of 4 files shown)


peteski22:
### Code review

Found 3 issues:

1. Data race on shared `Headers` map. `headerTransport.headers` aliases `Config.Headers` (exported, mutable) rather than copying it. If any caller mutates `cfg.Headers` after `HTTPClient()` returns while a concurrent request is in `RoundTrip` ranging over the map, that is an unsynchronised concurrent read/write. Fix by cloning at construction. (`.claude/CLAUDE.md` Go Proverbs: *"Don't communicate by sharing memory, share memory by communicating."*)

https://github.com/mozilla-ai/any-llm-go/blob/e9c9344222f1aee2aa51ccae645e6dde71c6c5b4/config/config.go#L184-L196

2. `WithHeader` placed out of alphabetical order — sits between `WithBaseURL` and `WithExtra`. Correct order: `WithAPIKey`, `WithBaseURL`, `WithExtra`, `WithHeader`, `WithHTTPClient`, `WithTimeout`. (`.claude/CLAUDE.md` Go Language Rules: *"Within each section, top-level declarations should be in alphabetical order"*. Root `CLAUDE.md` File Organization: *"Exported methods (alphabetically)"*.)

https://github.com/mozilla-ai/any-llm-go/blob/e9c9344222f1aee2aa51ccae645e6dde71c6c5b4/config/config.go#L96-L108

3. `HTTPClient()` docstring is now inaccurate. The note says *"If a custom client was provided via WithHTTPClient, that pointer is returned"* — but when `Headers` is non-empty, `HTTPClient()` constructs a new `*http.Client` wrapping the original, so the returned pointer is different. Update the doc to reflect the wrapping and the fact that `Timeout`/`CheckRedirect`/`Jar` are copied at that point.

https://github.com/mozilla-ai/any-llm-go/blob/e9c9344222f1aee2aa51ccae645e6dde71c6c5b4/config/config.go#L172-L199

config/config.go
```diff
@@ -155,11 +180,43 @@ func (c *Config) HTTPClient() *http.Client {
...
 
+// headerTransport is an http.RoundTripper that injects headers on every request.
+type headerTransport struct {
+	base    http.RoundTripper
+	headers map[string]string
+}
+
+func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	for key, value := range t.headers {
+		req.Header.Set(key, value)
```
peteski22: `RoundTrip` mutates the caller's request. `net/http` contract: *"RoundTrip should not modify the request, except for consuming and closing the Body."* Gateway provider already does this correctly — clone first:

```go
clone := req.Clone(req.Context())
for k, v := range t.headers {
    clone.Header.Set(k, v)
}
return t.base.RoundTrip(clone)
```

See `providers/gateway/gateway.go:373-378` for the established pattern.

providers/anthropic/anthropic.go
```diff
@@ -112,6 +112,7 @@ func New(opts ...config.Option) (*Provider, error) {
 
 	clientOpts := []option.RequestOption{
 		option.WithAPIKey(apiKey),
+		option.WithHTTPClient(cfg.HTTPClient()),
```
peteski22, line 121: Correct fix, but silent behavior change. Before this PR the Anthropic SDK used its own default HTTP client (no timeout). Now `cfg.Timeout` applies — existing callers with Anthropic calls >120s (default) will start timing out. Please note in PR description + CHANGELOG.

peteski22 (commented):
Inline comments cover line-anchored issues. Non-anchored items:

**Tests**
- No end-to-end test. Spin up `httptest.NewServer`, assert inbound headers match. Catches the `RoundTrip` mutation bug and real injection regressions.
- Missing test for duplicate-key semantics (`WithHeader("X","a"); WithHeader("X","b")`). Documents overwrite vs append behavior — required if `Headers` switches to `http.Header`.

**Follow-ups (not blocking)**
- Gateway provider on `main` has its own `headerTransport` (`providers/gateway/gateway.go:125-128,373-378`). Migrate to `config.WithHeader` in a follow-up to dedupe.
- Platform provider builds its own `&http.Client{Timeout: 30s}` and won't pick up `WithHeader`. Out of scope per gateway replacement.

Verified against e9c9344 + origin/main. Extends prior comment.

config/config.go
```diff
@@ -18,6 +18,11 @@ type Config struct {
 	// BaseURL is the base URL for the API. If empty, the provider's default is used.
 	BaseURL string
 
+	// Headers holds extra HTTP headers to include on every request.
+	// Useful for proxy authentication (e.g., Cloudflare AI Gateway's
+	// cf-aig-authorization header).
+	Headers map[string]string
```
peteski22: `map[string]string` silently overwrites duplicate keys and can't represent multi-valued headers (RFC 7230 §3.2.2). Suggest `http.Header` to match stdlib, with `WithHeader` calling `Add` (append) or `Set` (overwrite) — pick one and document the semantic.

config/config.go
```diff
@@ -155,11 +180,43 @@ func (c *Config) HTTPClient() *http.Client {
...
+				Jar:           c.httpClient.Jar,
+			}
+		}
 	})
 
 	return c.httpClient
 }
 
+// headerTransport is an http.RoundTripper that injects headers on every request.
+type headerTransport struct {
```
peteski22: Type belongs in the types section near `Config`/`Option`, not after `HTTPClient()`. Project `CLAUDE.md` File Organization: *"Types (exported first, then unexported helpers)"* precede constructor and methods.

config/config.go
```diff
@@ -155,11 +180,43 @@ func (c *Config) HTTPClient() *http.Client {
 		if c.httpClient == nil {
 			c.httpClient = &http.Client{Timeout: c.Timeout}
 		}
+
+		// Wrap the transport to inject custom headers on every request.
```
peteski22: Narration — restates the next three lines. Drop, or replace with *why* (e.g. "applied transport-layer so header injection is provider-agnostic").

config/config_test.go
```diff
@@ -572,3 +572,126 @@ func TestHTTPClientCaching(t *testing.T) {
...
+			t.Parallel()
+
+			cfg, err := New(WithHeader(tc.key, tc.value))
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, cfg.Headers)
```
peteski22, line 633: Only checks `Headers != nil`. Doesn't verify the trimmed key or value landed. Add:

```go
require.Equal(t, expectedKey, /* resolved key */)
require.Equal(t, tc.value, cfg.Headers[expectedKey])
```

config/config_test.go
```diff
@@ -572,3 +572,126 @@ func TestHTTPClientCaching(t *testing.T) {
...
+	t.Parallel()
+
+	cfg, err := New()
+	require.NoError(t, err)
+
+	client := cfg.HTTPClient()
+	require.NotNil(t, client)
+
+	// Without headers, transport should NOT be a headerTransport.
+	_, ok := client.Transport.(*headerTransport)
```
peteski22: Passes for the wrong reason. With no headers + no custom client, `client.Transport` is `nil`. Type assertion on `nil` returns `ok=false`, so this succeeds even if the wrapping logic were broken. Tighten:

```go
require.Nil(t, client.Transport)
```

jrschumacher:
@peteski22 Sounds good I'll take a look at these issues.

providers/anthropic/anthropic.go
peteski22: Documented on `config.WithTimeout` and called out in the PR summary. The underlying design — a whole-request `http.Client.Timeout` that caps streaming responses across all providers — is tracked in #152.

peteski22 (approved):

PR #80, opened 2026-04-10
fix(gemini): pass json_schema payload in structured output requests

### LINKED ISSUE #79: [BUG] Gemini provider ignores json_schema response format — schema not sent to API
## Description

When using `ResponseFormat` with `Type: "json_schema"` on the Gemini provider, the schema is silently dropped. The provider sets `ResponseMIMEType = "application/json"` but never populates `ResponseJsonSchema`, so Gemini returns an unstructured prose response instead of one conforming to the schema.

The [Gemini structured output docs](https://ai.google.dev/gemini-api/docs/structured-output) require both fields to be set for schema-constrained output.

## Steps to Reproduce

1. Create a Gemini provider with a valid API key
2. Call `Completion` with `ResponseFormat.Type = "json_schema"` and a non-nil `JSONSchema` with a schema map
3. Observe the response — it is unstructured prose, not JSON conforming to the schema

## Expected Behavior

The response content should be a JSON string conforming to the provided schema, e.g.:

```json
{
  "name": "Paris",
  "population": 2161000,
  "country": "France"
}
```

## Actual Behavior

An unstructured prose response is returned, e.g.:

```
Paris is the capital and most populous city of France, with a population of about 2.1 million...
```

`cfg.ResponseJsonSchema` is never set in `applyResponseFormat` (`providers/gemini/gemini.go`), so Gemini receives no schema constraint.

## Environment

- Go version: go1.25+
- OS: Linux / macOS
- any-llm-go version: v0.9.0
- Provider(s): gemini

## Code Example

```go
package main

import (
	"context"
	"fmt"
	"log"

	anyllm "github.com/mozilla-ai/any-llm-go"
	"github.com/mozilla-ai/any-llm-go/providers/gemini"
)

func main() {
	ctx := context.Background()

	// Requires GEMINI_API_KEY or GOOGLE_API_KEY to be set.
	provider, err := gemini.New()
	if err != nil {
		log.Fatal(err)
	}

	schema := map[string]any{
		"type": "object",
		"properties": map[string]any{
			"name":       map[string]any{"type": "string"},
			"population": map[string]any{"type": "integer"},
			"country":    map[string]any{"type": "string"},
		},
		"required": []string{"name", "population", "country"},
	}

	strict := true

	resp, err := provider.Completion(ctx, anyllm.CompletionParams{
		Model: "gemini-3.1-flash-lite-preview",
		Messages: []anyllm.Message{
			{Role: anyllm.RoleUser, Content: "Give me information about Paris."},
		},
		ResponseFormat: &anyllm.ResponseFormat{
			Type: "json_schema",
			JSONSchema: &anyllm.JSONSchema{
				Name:   "city_info",
				Schema: schema,
				Strict: &strict,
			},
		},
	})
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(resp.Choices[0].Message.Content)
}
```

## Additional Context

The root cause is in `applyResponseFormat` (`providers/gemini/gemini.go`). The function only handled the `json_object` case—the `json_schema` case was missing entirely:

```go
// current code — json_schema branch absent
func applyResponseFormat(cfg *genai.GenerateContentConfig, format *providers.ResponseFormat) {
	if format.Type == responseFormatJSON {
		cfg.ResponseMIMEType = responseMIMETypeJSON
	}
}
```

The fix requires adding a `json_schema` branch that sets both `ResponseMIMEType` and `ResponseJsonSchema`.

The OpenAI-compatible provider (`providers/openai/compatible.go`) already handles this correctly via `convertResponseFormat`.

apikdech:
## Summary
Fixes a bug where the Gemini provider silently dropped the JSON schema when `ResponseFormat.Type = "json_schema"` was set. `applyResponseFormat` only handled the `json_object` case, leaving `cfg.ResponseJsonSchema` unpopulated.

Gemini requires both `ResponseMIMEType` and `ResponseJsonSchema` to be set for schema-constrained output—only the MIME type was being set, so the API returned unstructured prose instead of schema-conforming JSON.

Closes #79 

## Changes
- Add `responseFormatJSONSchema = "json_schema"` constant to `providers/gemini/gemini.go`
- Add `json_schema` branch in `applyResponseFormat` that sets both `cfg.ResponseMIMEType` and `cfg.ResponseJsonSchema`
- Add nil guard at the top of `applyResponseFormat` (defensive, consistent with other providers)
- Add `TestConvertParams` to verify `json_object`, `json_schema`, and nil response formats are wired through `convertParams` correctly
- Extend `TestApplyResponseFormat` with three new sub-cases covering the `json_schema` branch and its nil guards

## Testing
Unit tests (no API key required):

```bash
make test-unit
```

The code example from #79  can also be used to confirm the response is now a JSON object conforming to the schema instead of unstructured prose.

## Checklist
- [ ] Documentation updated (if needed)

### DIFF
--- a/providers/gemini/gemini.go
+++ b/providers/gemini/gemini.go
@@ -58,10 +58,11 @@ const (
 
 // Response format and tool type constants.
 const (
-	responseMIMETypeJSON = "application/json"
-	responseFormatJSON   = "json_object"
-	toolCallFallbackName = "function"
-	toolCallType         = "function"
+	responseMIMETypeJSON     = "application/json"
+	responseFormatJSON       = "json_object"
+	responseFormatJSONSchema = "json_schema"
+	toolCallFallbackName     = "function"
+	toolCallType             = "function"
 )
 
 // ID prefix constants for generated identifiers.
@@ -489,8 +490,20 @@ func (s *streamState) processResponse(resp *genai.GenerateContentResponse) ([]pr
 }
 
 // applyResponseFormat configures the response format on the config.
+// For json_schema, Gemini requires both responseMimeType application/json and
+// responseJsonSchema (see https://ai.google.dev/gemini-api/docs/structured-output).
 func applyResponseFormat(cfg *genai.GenerateContentConfig, format *providers.ResponseFormat) {
-	if format.Type == responseFormatJSON {
+	if format == nil {
+		return
+	}
+	switch format.Type {
+	case responseFormatJSONSchema:
+		if format.JSONSchema == nil || format.JSONSchema.Schema == nil {
+			return
+		}
+		cfg.ResponseMIMEType = responseMIMETypeJSON
+		cfg.ResponseJsonSchema = format.JSONSchema.Schema
+	case responseFormatJSON:
 		cfg.ResponseMIMEType = responseMIMETypeJSON
 	}
 }
--- a/providers/gemini/gemini_test.go
+++ b/providers/gemini/gemini_test.go
@@ -1194,6 +1194,68 @@ func TestStreamStateFinalChunk(t *testing.T) {
 	})
 }
 
+func TestConvertParams(t *testing.T) {
+	t.Parallel()
+
+	provider, err := New(config.WithAPIKey("test-key"))
+	require.NoError(t, err)
+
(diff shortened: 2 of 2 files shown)


tbille (approved):
amazing, thanks for the contribution!

PR #81, opened 2026-04-13
feat(gateway): add gateway provider with platform and non-platform mode support

tbille:
## Summary

- Adds a new `providers/gateway` package that connects to an any-llm gateway server, superseding `providers/platform`. Where the platform provider decrypts provider keys client-side and fans out to individual provider SDKs, the gateway delegates everything to a single server-side endpoint.
- Supports **platform mode** (Bearer auth) and **non-platform mode** (`X-AnyLLM-Key` header)
- Adds `InsufficientFundsError` (402) to the shared `errors` package; gateway-specific error types `UpstreamProviderError` (502) and `TimeoutError` (504) live in the gateway package
- Extends `openai.CompatibleConfig` with a `RequireBaseURL` flag so providers with no sensible default URL (like the gateway) can fail fast with a clear, env-var-aware error at construction time

### Gateway Provider (`providers/gateway/`)

The gateway provider acts as a thin wrapper around `openai.CompatibleProvider`, pointing the OpenAI SDK at a gateway URL. Two authentication modes are supported:

**Platform mode** (most common): The platform token is used as standard Bearer authentication via the OpenAI SDK's API key mechanism.

```go
provider, err := gateway.New(
    config.WithBaseURL("http://localhost:8000/v1"),
    config.WithAPIKey("tk_your_platform_token"),
    gateway.WithPlatformMode(),
)
```

**Non-platform mode**: A gateway API key is sent via the `X-AnyLLM-Key` header using a custom `http.RoundTripper` wrapper, requiring no changes to shared code.

```go
provider, err := gateway.New(
    config.WithBaseURL("http://localhost:8000/v1"),
    gateway.WithGatewayKey("your-gateway-key"),
)
```

Platform mode auto-detection: If `GATEWAY_PLATFORM_TOKEN` is set and no explicit API key or gateway key is provided, platform mode is automatically enabled.

| Env Var | Purpose |
|---------|---------|
| `GATEWAY_API_BASE` | Gateway URL (e.g. `http://localhost:8000/v1`) |
| `GATEWAY_API_KEY` | Non-platform mode key (sent via `X-AnyLLM-Key` header) |
| `GATEWAY_PLATFORM_TOKEN` | Platform mode token (sent as Bearer auth) |

### Required Base URL (`providers/openai/compatible.go`)

`CompatibleConfig` gains a `RequireBaseURL bool` field, parallel to the existing `RequireAPIKey`. When set, `NewCompatible` enforces a non-empty base URL and returns a human-readable error that names the configured env var. The gateway provider sets this flag — there is no sensible default gateway URL — so misconfiguration fails fast at construction time rather than surfacing later as an opaque network error.

### Relationship to `providers/platform`

The gateway provider replaces the platform provider for new integrations. The platform provider (`providers/platform/`) authenticates with the platform API to decrypt provider keys, then fans out to individual provider SDKs (OpenAI, Anthropic, etc.) client-side. The gateway provider instead sends all requests through a single server-side gateway endpoint, which handles provider routing, key management, and usage tracking.

`providers/platform` remains in the repo for backward compatibility but new consumers should use `providers/gateway`.

### Error Types

- `InsufficientFundsError` / `ErrInsufficientFunds` (HTTP 402) — shared `errors` package
- `UpstreamProviderError` / `ErrUpstreamProvider` (HTTP 502) — `providers/gateway` package
- `TimeoutError` / `ErrTimeout` (HTTP 504) — `providers/gateway` package

Gateway-specific error types live in the gateway package since upstream-provider and gateway-timeout are gateway domain concepts. `InsufficientFundsError` is in the shared package since billing errors can apply to any provider.

### Post-review cleanup

Craftsmanship work applied on top of the initial implementation:

- `gateway.New()` no longer duplicates base-URL resolution. User opts flow straight through; auth-specific opts are layered via `slices.Clone + append`. The HTTPClient-drop regression that recurred across earlier review rounds is structurally impossible in this form.
- Package-name stutter removed: `GatewayTimeoutError` → `TimeoutError`, `ErrGatewayTimeout` → `ErrTimeout`.
- Internal names clarified: `defaultNonPlatformKey` → `placeholderAPIKey` (with a godoc explaining why the SDK needs *any* key when auth goes via a custom header), `gatewayHeader` → `apiKeyHeaderName`.
- Single-use error constructors inlined into `ConvertError`; every package constant and exported sentinel now carries a godoc line; type ordering corrected (exported before unexported helpers).
- `errors.NewBaseError` renamed to `errors.New` with an explanatory godoc. The old name read like "construct an error of kind `BaseError`" — but `BaseError` is never returned directly; the helper exists only so packages outside `errors` can populate the unexported `sentinel` field that drives `errors.Is` matching.

### Tests

- Unit tests: constructor validation, mode detection, header injection, completion/streaming, error conversion for all gateway-specific status codes (402, 502, 504) plus inherited ones (401, 404, 429)
- `TestNewCompatibleRequireBaseURL` — six cases asserting the full error message on both branches (env var named vs not)
- `TestExtraValueHandling` — table-driven with six cases, using `httptest` to assert wire-level behaviour: silent fallthrough on wrong-typed `Extra` values, empty-string handling, env-var fallthrough, and proof-of-mode via which auth header is actually sent
- `TestPlatformModeDetection` — reorganised to table-driven, scoped to auto-detect-suppression cases
- Integration tests: guarded by `GATEWAY_API_BASE` and `GATEWAY_PLATFORM_TOKEN` env vars, verified against a live gateway
- All existing tests continue to pass

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -51,6 +51,30 @@ type BaseError struct {
 	sentinel error
 }
 
+// New constructs a BaseError for embedding in a caller-defined error type.
+// It exists so packages outside this one can populate the unexported
+// sentinel field that drives errors.Is matching; without it, external
+// packages could only set the exported fields via struct literals and
+// errors.Is(err, ErrFoo) would never match.
+//
+// Typical usage from a provider package:
+//
+//	type UpstreamProviderError struct{ errors.BaseError }
+//
+//	func NewUpstreamProviderError(provider string, err error) *UpstreamProviderError {
+//	    return &UpstreamProviderError{
+//	        BaseError: errors.New("upstream_provider", provider, err, ErrUpstreamProvider),
+//	    }
+//	}
+func New(code string, provider string, err error, sentinel error) BaseError {
+	return BaseError{
+		Code:     code,
+		Provider: provider,
+		Err:      err,
+		sentinel: sentinel,
+	}
+}
+
 // Error implements the error interface.
 func (e *BaseError) Error() string {
 	msg := ""
@@ -153,17 +177,6 @@ func NewAuthenticationError(provider string, err error) *AuthenticationError {
 			Err:      err,
 			sentinel: ErrAuthentication,
 		},
-	}
-}
-
-// NewBaseError creates a BaseError with the given fields. This allows provider
-// packages to define their own error types that embed BaseError with a sentinel.
-func NewBaseError(code, provider string, err, sentinel error) BaseError {
-	return BaseError{
-		Code:     code,
-		Provider: provider,
-		Err:      err,
-		sentinel: sentinel,
 	}
 }
 
@@ -271,6 +284,11 @@ func NewUnsupportedParamError(provider string, param string) *UnsupportedParamEr
 // NewInsufficientFundsError creates a new InsufficientFundsError.
 func NewInsufficientFundsError(provider string, err error) *InsufficientFundsError {
 	return &InsufficientFundsError{
-		BaseError: NewBaseError(CodeInsufficientFunds, provider, err, ErrInsufficientFunds),
+		BaseError: BaseError{
+			Code:     CodeInsufficientFunds,
+			Provider: provider,
+			Err:      err,
+			sentinel: ErrInsufficientFunds,
+		},
 	}
 }
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,8 +18,6 @@ const (
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
 	CodeInsufficientFunds   = "insufficient_funds"
-	CodeUpstreamProvider    = "upstream_provider"
-	CodeGatewayTimeout      = "gateway_timeout"
 )
 
 // Sentinel errors for type checking with errors.Is().
@@ -35,8 +33,6 @@ var (
 	ErrUnsupportedProvider = stderrors.New("unsupported provider")
 	ErrUnsupportedParam    = stderrors.New("unsupported parameter")
 	ErrInsufficientFunds   = stderrors.New("insufficient funds")
-	ErrUpstreamProvider    = stderrors.New("upstream provider error")
-	ErrGatewayTimeout      = stderrors.New("gateway timeout")
 )
 
 // BaseError is the base error type for all any-llm errors.
@@ -133,16 +129,6 @@ type UnsupportedParamError struct {
 
 // InsufficientFundsError is returned when the account has insufficient funds (HTTP 402).
 type InsufficientFundsError struct {
-	BaseError
-}
-
-// UpstreamProviderError is returned when the upstream provider is unreachable (HTTP 502).
-type UpstreamProviderError struct {
-	BaseError
-}
-
-// GatewayTimeoutError is returned when the gateway times out (HTTP 504).
-type GatewayTimeoutError struct {
 	BaseError
 }
 
@@ -271,38 +257,20 @@ func NewUnsupportedParamError(provider string, param string) *UnsupportedParamEr
 	}
 }
 
+// NewBaseError creates a BaseError with the given fields. This allows provider
+// packages to define their own error types that embed BaseError with a sentinel.
+func NewBaseError(code, provider string, err, sentinel error) BaseError {
+	return BaseError{
+		Code:     code,
+		Provider: provider,
+		Err:      err,
+		sentinel: sentinel,
+	}
+}
+
 // NewInsufficientFundsError creates a new InsufficientFundsError.
 func NewInsufficientFundsError(provider string, err error) *InsufficientFundsError {
 	return &InsufficientFundsError{
-		BaseError: BaseError{
-			Code:     CodeInsufficientFunds,
-			Provider: provider,
-			Err:      err,
-			sentinel: ErrInsufficientFunds,
-		},
-	}
-}
-
-// NewUpstreamProviderError creates a new UpstreamProviderError.
-func NewUpstreamProviderError(provider string, err error) *UpstreamProviderError {
-	return &UpstreamProviderError{
-		BaseError: BaseError{
-			Code:     CodeUpstreamProvider,
-			Provider: provider,
-			Err:      err,
-			sentinel: ErrUpstreamProvider,
-		},
-	}
-}
-
-// NewGatewayTimeoutError creates a new GatewayTimeoutError.
-func NewGatewayTimeoutError(provider string, err error) *GatewayTimeoutError {
-	return &GatewayTimeoutError{
-		BaseError: BaseError{
-			Code:     CodeGatewayTimeout,
-			Provider: provider,
-			Err:      err,
-			sentinel: ErrGatewayTimeout,
-		},
+		BaseError: NewBaseError(CodeInsufficientFunds, provider, err, ErrInsufficientFunds),
 	}
 }
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -17,6 +17,9 @@ const (
 	CodeMissingAPIKey       = "missing_api_key"
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
+	CodeInsufficientFunds   = "insufficient_funds"
+	CodeUpstreamProvider    = "upstream_provider"
+	CodeGatewayTimeout      = "gateway_timeout"
 )
 
 // Sentinel errors for type checking with errors.Is().
@@ -31,6 +34,9 @@ var (
 	ErrMissingAPIKey       = stderrors.New("missing API key")
 	ErrUnsupportedProvider = stderrors.New("unsupported provider")
 	ErrUnsupportedParam    = stderrors.New("unsupported parameter")
+	ErrInsufficientFunds   = stderrors.New("insufficient funds")
+	ErrUpstreamProvider    = stderrors.New("upstream provider error")
+	ErrGatewayTimeout      = stderrors.New("gateway timeout")
 )
 
 // BaseError is the base error type for all any-llm errors.
@@ -123,6 +129,21 @@ type UnsupportedProviderError struct {
 type UnsupportedParamError struct {
 	BaseError
 	Param string // The unsupported parameter name
+}
+
+// InsufficientFundsError is returned when the account has insufficient funds (HTTP 402).
+type InsufficientFundsError struct {
+	BaseError
+}
+
+// UpstreamProviderError is returned when the upstream provider is unreachable (HTTP 502).
+type UpstreamProviderError struct {
+	BaseError
+}
+
+// GatewayTimeoutError is returned when the gateway times out (HTTP 504).
+type GatewayTimeoutError struct {
+	BaseError
 }
 
 // NewRateLimitError creates a new RateLimitError.
@@ -247,5 +268,41 @@ func NewUnsupportedParamError(provider string, param string) *UnsupportedParamEr
 			sentinel: ErrUnsupportedParam,
 		},
 		Param: param,
+	}
+}
+
+// NewInsufficientFundsError creates a new InsufficientFundsError.
+func NewInsufficientFundsError(provider string, err error) *InsufficientFundsError {
+	return &InsufficientFundsError{
+		BaseError: BaseError{
+			Code:     CodeInsufficientFunds,
+			Provider: provider,
+			Err:      err,
+			sentinel: ErrInsufficientFunds,
+		},
+	}
+}
+
+// NewUpstreamProviderError creates a new UpstreamProviderError.
+func NewUpstreamProviderError(provider string, err error) *UpstreamProviderError {
+	return &UpstreamProviderError{
+		BaseError: BaseError{
+			Code:     CodeUpstreamProvider,
+			Provider: provider,
+			Err:      err,
+			sentinel: ErrUpstreamProvider,
+		},
+	}
+}
+
+// NewGatewayTimeoutError creates a new GatewayTimeoutError.
+func NewGatewayTimeoutError(provider string, err error) *GatewayTimeoutError {
+	return &GatewayTimeoutError{
+		BaseError: BaseError{
+			Code:     CodeGatewayTimeout,
+			Provider: provider,
+			Err:      err,
+			sentinel: ErrGatewayTimeout,
+		},
 	}
 }
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -156,6 +156,17 @@ func NewAuthenticationError(provider string, err error) *AuthenticationError {
 	}
 }
 
+// NewBaseError creates a BaseError with the given fields. This allows provider
+// packages to define their own error types that embed BaseError with a sentinel.
+func NewBaseError(code, provider string, err, sentinel error) BaseError {
+	return BaseError{
+		Code:     code,
+		Provider: provider,
+		Err:      err,
+		sentinel: sentinel,
+	}
+}
+
 // NewInvalidRequestError creates a new InvalidRequestError.
 func NewInvalidRequestError(provider string, err error) *InvalidRequestError {
 	return &InvalidRequestError{
@@ -254,17 +265,6 @@ func NewUnsupportedParamError(provider string, param string) *UnsupportedParamEr
 			sentinel: ErrUnsupportedParam,
 		},
 		Param: param,
-	}
-}
-
-// NewBaseError creates a BaseError with the given fields. This allows provider
-// packages to define their own error types that embed BaseError with a sentinel.
-func NewBaseError(code, provider string, err, sentinel error) BaseError {
-	return BaseError{
-		Code:     code,
-		Provider: provider,
-		Err:      err,
-		sentinel: sentinel,
 	}
 }
 
--- a/providers/gateway/gateway.go
+++ b/providers/gateway/gateway.go
@@ -4,7 +4,7 @@
 //
 // The gateway supports two authentication modes:
 //   - Platform mode: uses a platform token as standard Bearer auth
-//   - Non-platform mode: sends a gateway API key via the X-AnyLLM-Key header
+//   - Non-platform mode: sends a gateway API key via the gatewayHeader header
 package gateway
 
 import (
@@ -23,11 +23,15 @@ import (
 
 // Provider configuration constants.
 const (
-	envAPIBase       = "GATEWAY_API_BASE"
-	envAPIKey        = "GATEWAY_API_KEY"
-	envPlatformToken = "GATEWAY_PLATFORM_TOKEN"
-	gatewayHeader    = "X-AnyLLM-Key"
-	providerName     = "gateway"
+	bearerPrefix          = "Bearer "
+	defaultNonPlatformKey = "gateway-no-key"
+	envAPIBase            = "GATEWAY_API_BASE"
+	envAPIKey             = "GATEWAY_API_KEY"
+	envPlatformToken      = "GATEWAY_PLATFORM_TOKEN"
+	extraKeyGatewayKey    = "gateway_key"
+	extraKeyPlatformMode  = "platform_mode"
+	gatewayHeader         = "X-AnyLLM-Key"
+	providerName          = "gateway"
 )
 
 // Ensure Provider implements the required interfaces.
@@ -47,6 +51,22 @@ type Provider struct {
 	platformMode bool
 }
 
+// headerTransport wraps an http.RoundTripper to inject custom headers into
+// every request. Used in non-platform mode to add the gateway key header.
+type headerTransport struct {
+	base   http.RoundTripper
+	header string
+	value  string
+}
+
+// RoundTrip implements http.RoundTripper by cloning the request and injecting
+// the configured header before delegating to the base transport.
+func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	clone := req.Clone(req.Context())
+	clone.Header.Set(t.header, t.value)
+	return t.base.RoundTrip(clone)
+}
+
 // New creates a new gateway provider.
 //
 // The gateway base URL is required and can be set via config.WithBaseURL() or
@@ -58,7 +78,7 @@ type Provider struct {
 //   - If GATEWAY_PLATFORM_TOKEN is set and no explicit API key is provided,
 //     platform mode is auto-detected.
 //   - Otherwise, non-platform mode is used with the key from WithGatewayKey()
-//     or GATEWAY_API_KEY, sent via the X-AnyLLM-Key header.
+//     or GATEWAY_API_KEY, sent via the gatewayHeader header.
 func New(opts ...config.Option) (*Provider, error) {
 	cfg, err := config.New(opts...)
 	if err != nil {
@@ -76,44 +96,61 @@ func New(opts ...config.Option) (*Provider, error) {
 	}
 
 	platformMode, platformToken := resolvePlatformMode(cfg)
-
-	if platformMode {
-		return newPlatformProvider(cfg, baseURL, platformToken)
+	if platformMode && platformToken == "" {
+		return nil, fmt.Errorf(
+			"platform mode requires a token (pass WithAPIKey option or set %s env var)",
+			envPlatformToken,
+		)
 	}
 
-	return newNonPlatformProvider(cfg, baseURL)
+	// Build options for the underlying OpenAI-compatible provider.
+	compatOpts := []config.Option{config.WithBaseURL(baseURL)}
+	if cfg.Timeout > 0 {
+		compatOpts = append(compatOpts, config.WithTimeout(cfg.Timeout))
+	}
+
+	httpClient := cfg.HTTPClient()
+	if platformMode {
+		compatOpts = append(compatOpts, config.WithAPIKey(platformToken))
+	} else {
+		gatewayKey := resolveGatewayKey(cfg)
+		if gatewayKey != "" {
+			httpClient = newHeaderClient(httpClient, bearerPrefix+gatewayKey)
+		}
+	}
+	compatOpts = append(compatOpts, config.WithHTTPClient(httpClient))
+
+	base, err := openaiProvider.NewCompatible(openaiProvider.CompatibleConfig{
+		APIKeyEnvVar:   "", // Gateway uses its own key resolution.
+		BaseURLEnvVar:  "", // Already resolved above.
+		Capabilities:   capabilities(),
+		DefaultAPIKey:  defaultNonPlatformKey, // Placeholder; non-platform doesn't need real auth.
+		DefaultBaseURL: "",                    // No default; base URL is required.
+		Name:           providerName,
+		RequireAPIKey:  false, // Gateway handles auth separately.
+	}, compatOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Provider{
+		CompatibleProvider: base,
+		platformMode:       platformMode,
+	}, nil
 }
 
 // WithGatewayKey sets the gateway API key for non-platform mode authentication.
-// The key is sent as a Bearer-formatted value in the X-AnyLLM-Key header
+// The key is sent as a Bearer-formatted value in the gatewayHeader header
 // (i.e., "X-AnyLLM-Key: Bearer <key>").
 func WithGatewayKey(key string) config.Option {
-	return config.WithExtra("gateway_key", key)
+	return config.WithExtra(extraKeyGatewayKey, key)
 }
 
 // WithPlatformMode explicitly enables platform mode authentication.
 // In platform mode, the token (from config.WithAPIKey or GATEWAY_PLATFORM_TOKEN)
 // is used as standard Bearer authentication.
 func WithPlatformMode() config.Option {
-	return config.WithExtra("platform_mode", true)
-}
-
-// ConvertError extends the base OpenAI error conversion with gateway-specific
-// HTTP status codes (402, 502, 504). This method handles raw (unconverted)
-// errors and implements the providers.ErrorConverter interface.
-func (p *Provider) ConvertError(err error) error {
-	if err == nil {
-		return nil
-	}
-
-	var apiErr *openai.Error
-	if stderrors.As(err, &apiErr) {
-		if converted := convertGatewayError(apiErr); converted != nil {
-			return converted
-		}
-	}
-
-	return p.CompatibleProvider.ConvertError(err)
+	return config.WithExtra(extraKeyPlatformMode, true)
 }
 
 // Completion performs a chat completion request.
@@ -124,7 +161,7 @@ func (p *Provider) Completion(
 ) (*providers.ChatCompletion, error) {
 	resp, err := p.CompatibleProvider.Completion(ctx, params)
 	if err != nil {
-		return nil, p.reclassifyError(err)
+		return nil, p.ConvertError(err)
 	}
 
 	return resp, nil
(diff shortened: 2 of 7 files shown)


providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	return newNonPlatformProvider(cfg, baseURL)
+}
+
+// WithGatewayKey sets the gateway API key for non-platform mode authentication.
+// The key is sent via the X-AnyLLM-Key header.
+func WithGatewayKey(key string) config.Option {
+	return config.WithExtra("gateway_key", key)
+}
```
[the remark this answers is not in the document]
tbille: Fixed: Clarified the doc comment to explicitly state that the key is sent as a Bearer-formatted value in the header (i.e., `X-AnyLLM-Key: Bearer <key>`).

providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+// Completion performs a chat completion request.
+// In platform mode, gateway-specific errors are converted to typed errors.
+func (p *Provider) Completion(
+	ctx context.Context,
+	params providers.CompletionParams,
+) (*providers.ChatCompletion, error) {
+	resp, err := p.CompatibleProvider.Completion(ctx, params)
+	if err != nil && p.platformMode {
+		return nil, p.ConvertError(err)
+	}
```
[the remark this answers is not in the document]
tbille: Fixed: Gateway-specific error conversion (402/502/504) now applies in both platform and non-platform modes. The `Completion`, `CompletionStream`, `Embedding`, and `ListModels` methods all call `reclassifyError` regardless of mode.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+	if gatewayKey != "" {
+		httpClient := &http.Client{
+			Timeout: cfg.Timeout,
+			Transport: &headerTransport{
+				base:   http.DefaultTransport,
+				header: gatewayHeader,
+				value:  "Bearer " + gatewayKey,
+			},
+		}
+		compatOpts = append(compatOpts, config.WithHTTPClient(httpClient))
```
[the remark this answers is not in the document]
tbille: Fixed: Now starts from `cfg.HTTPClient().Transport` as the base round-tripper, preserving any user-supplied transport settings (proxies, mTLS, tracing, etc.).

providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+
+	base, err := openaiProvider.NewCompatible(openaiProvider.CompatibleConfig{
+		APIKeyEnvVar:   "",
+		BaseURLEnvVar:  "",
+		Capabilities:   capabilities(),
+		DefaultAPIKey:  "",
+		DefaultBaseURL: "",
+		Name:           providerName,
+		RequireAPIKey:  false,
+	}, config.WithAPIKey(token), config.WithBaseURL(baseURL))
```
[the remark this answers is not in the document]
tbille: Fixed: Added `forwardConfigOptions` helper that forwards base URL and timeout to the underlying CompatibleProvider. Both `newPlatformProvider` and `newNonPlatformProvider` now use it.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+			if token == "" {
+				token = cfg.ResolveEnv(envPlatformToken)
+			}
+			return true, token
+		}
+	}
+
+	// Auto-detect: GATEWAY_PLATFORM_TOKEN set and no explicit API key.
+	platformToken := cfg.ResolveEnv(envPlatformToken)
+	if platformToken != "" && cfg.APIKey == "" {
```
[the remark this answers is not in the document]
tbille: Fixed: Auto-detection now checks for the presence of a gateway key as well. If `WithGatewayKey` or `GATEWAY_API_KEY` is configured, platform mode is not auto-enabled even when `GATEWAY_PLATFORM_TOKEN` is present.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,339 @@
...
+// ConvertError extends the base OpenAI error conversion with gateway-specific
+// HTTP status codes (402, 502, 504).
+func (p *Provider) ConvertError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	var apiErr *openai.Error
+	if stderrors.As(err, &apiErr) {
+		if converted := convertGatewayError(apiErr, err); converted != nil {
```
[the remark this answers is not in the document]
tbille: Fixed: Introduced `reclassifyError` for already-converted errors (used in Completion/Stream/Embedding/ListModels) that wraps the raw `*openai.Error` directly. `ConvertError` (for the ErrorConverter interface on raw errors) also wraps `apiErr` directly. This avoids the nested `insufficient_funds: provider_error: ...` pattern.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+		CompatibleProvider: base,
+		platformMode:       true,
+	}, nil
+}
+
+// forwardConfigOptions builds a slice of config options that forwards
+// user-supplied settings (base URL, timeout, HTTP client) to the underlying
+// CompatibleProvider. This ensures settings like WithTimeout and
+// WithHTTPClient are not silently dropped.
+func forwardConfigOptions(cfg *config.Config, baseURL string) []config.Option {
```
peteski22: `forwardConfigOptions` only forwards `BaseURL` and `Timeout`. If a user passes `config.WithHTTPClient(myClient)` to `gateway.New()`, the custom client is stored in the initial `cfg` but never forwarded to the underlying `CompatibleProvider`.

- In **platform mode**: the HTTPClient is always lost.
- In **non-platform mode with a gateway key**: the transport is partially preserved (used as the base for `headerTransport`), but other `http.Client` properties (`CheckRedirect`, `Jar`, client-level `Timeout`) are discarded.
- In **non-platform mode without a gateway key**: the HTTPClient is completely lost.

This is **newly introduced**: every other compatible-based provider (deepseek, groq, mistral, llamafile, llamacpp) passes `opts...` straight through to `NewCompatible`, so the client flows through naturally.

A user configuring a custom HTTP client for proxy settings, TLS configuration, or custom redirect policy would have that configuration silently ignored.

**Missing tests**: There is no test that verifies a custom HTTP client survives construction. A table-driven constructor test should cover:

| Mode         | WithHTTPClient | WithTimeout | WithGatewayKey | Expected                                         |
| ------------ | -------------- | ----------- | -------------- | ------------------------------------------------ |
| platform     | yes            | yes         | n/a            | client and timeout forwarded                     |
| platform     | no             | no          | n/a            | defaults used                                    |
| non-platform | yes            | yes         | yes            | client transport used as base, timeout forwarded |
| non-platform | yes            | no          | no             | client forwarded as-is                           |
| non-platform | no             | no          | yes            | default transport wrapped                        |

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+// the GATEWAY_API_BASE environment variable.
+//
+// Authentication mode is determined as follows:
+//   - If WithPlatformMode() is passed, platform mode is used. The token is
+//     resolved from config.WithAPIKey() or GATEWAY_PLATFORM_TOKEN.
+//   - If GATEWAY_PLATFORM_TOKEN is set and no explicit API key is provided,
+//     platform mode is auto-detected.
+//   - Otherwise, non-platform mode is used with the key from WithGatewayKey()
+//     or GATEWAY_API_KEY, sent via the X-AnyLLM-Key header.
+func New(opts ...config.Option) (*Provider, error) {
```
peteski22, line 146: `New()` calls `config.New(opts...)` to inspect the config, then manually resolves the base URL, determines platform mode, and resolves gateway keys. It then reconstructs a *new* set of options via `forwardConfigOptions` to pass to `NewCompatible`. This two-stage approach is unique to this provider and is the root cause of the HTTPClient bug mentioned re: `forwardConfigOptions`.

Every other compatible-based provider passes `opts...` straight through to `NewCompatible`. The `gateway` is fighting the `NewCompatible` abstraction rather than working with it.

Specific simplifications:

- **`BaseURLEnvVar`**: Set `BaseURLEnvVar: envAPIBase` on `CompatibleConfig` and let `NewCompatible` resolve it automatically, instead of resolving it manually in `New()`.
- **Validation**: If base URL is required, validate *after* `NewCompatible` returns rather than before. This keeps the opts passthrough clean and avoids the two-stage parsing entirely.
- **`resolvePlatformMode` and `resolveGatewayKey`**: Both are called exactly once from `New()`. They're shallow functions; narrow interfaces wrapping trivial logic, that add indirection without reducing complexity. Inline them into the constructor.

> *Go Proverbs — "Clear is better than clever."* A constructor you can read top-to-bottom is clearer than one that delegates to four helper functions. The indirection doesn't reduce complexity; it distributes it.

> *Ousterhout — deep modules over shallow ones.* `resolvePlatformMode`, `resolveGatewayKey`, `forwardConfigOptions` are shallow modules: narrow interfaces hiding trivial logic. They increase the number of things a reader needs to understand without making any individual thing simpler.

**Missing tests**: The extracted functions have some test coverage (`TestResolvePlatformMode`, `TestResolveGatewayKey`) but they are not comprehensive table-driven suites. Whether the code stays decomposed or is inlined, the constructor's input space needs systematic coverage; the same test matrix applies regardless of code structure.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	return newNonPlatformProvider(cfg, baseURL)
+}
+
+// WithGatewayKey sets the gateway API key for non-platform mode authentication.
+// The key is sent as a Bearer-formatted value in the X-AnyLLM-Key header
+// (i.e., "X-AnyLLM-Key: Bearer <key>").
+func WithGatewayKey(key string) config.Option {
+	return config.WithExtra("gateway_key", key)
```
peteski22: The file has a constants block at the top but uses magic strings elsewhere:

- `"gateway_key"` (lines 91, 329): coordination key between `WithGatewayKey` and `resolveGatewayKey`
- `"platform_mode"` (lines 98, 342):  coordination key between `WithPlatformMode` and `resolvePlatformMode`
- `"Bearer "` (line 254): auth scheme prefix
- `"gateway-no-key"` (line 264): synthetic dummy API key

The first two are especially fragile: they're coordination points between writer and reader functions. A typo in either location silently breaks the coordination with no compile-time error. Named constants eliminate this class of bug entirely.

`"gateway-no-key"` as a constant with a name and comment would make clear it's intentional rather than looking like a hardcoded credential.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+// In platform mode, the token (from config.WithAPIKey or GATEWAY_PLATFORM_TOKEN)
+// is used as standard Bearer authentication.
+func WithPlatformMode() config.Option {
+	return config.WithExtra("platform_mode", true)
+}
+
+// ConvertError extends the base OpenAI error conversion with gateway-specific
+// HTTP status codes (402, 502, 504). This method handles raw (unconverted)
+// errors and implements the providers.ErrorConverter interface.
+func (p *Provider) ConvertError(err error) error {
```
peteski22: `ConvertError`/`reclassifyError`/`convertGatewayError` over-decomposed

Three functions to do one job: convert gateway-specific HTTP status codes (402, 502, 504) to typed errors.

The root cause is Go's embedding semantics: `p.CompatibleProvider.Completion()` dispatches to `CompatibleProvider.ConvertError`, not the gateway's override. So the gateway can't just override `ConvertError` and have it take effect. The response to that constraint is over-engineered:

- `convertGatewayError`: 10-line switch, returns `nil` to mean "not a gateway error" — an unusual Go pattern (a bool return or inlining would be clearer).
- `reclassifyError`: 10 lines, extracts `*openai.Error` from the chain, calls `convertGatewayError`, falls through if nil.
- `ConvertError`: does the same as `reclassifyError` but falls through to `p.CompatibleProvider.ConvertError` instead of returning unchanged.

Suggestion: Collapse all three into `ConvertError`. It can handle both raw and already-converted errors by detecting already-converted errors and passing them through:

```go
func (p *Provider) ConvertError(err error) error {
    var apiErr *openai.Error
    if stderrors.As(err, &apiErr) {
        switch apiErr.StatusCode {
        case http.StatusPaymentRequired:
            return errors.NewInsufficientFundsError(providerName, apiErr)
        case http.StatusBadGateway:
            return errors.NewUpstreamProviderError(providerName, apiErr)
        case http.StatusGatewayTimeout:
            return errors.NewGatewayTimeoutError(providerName, apiErr)
        }
    }
    var baseErr *errors.BaseError
    if stderrors.As(err, &baseErr) {
        return err
    }
    return p.CompatibleProvider.ConvertError(err)
}
```

Then every method override calls `p.ConvertError(err)` and both `reclassifyError` and `convertGatewayError` are deleted.

> *Go Proverbs — "Clear is better than clever."* Three functions with subtle contracts (raw vs already-converted, nil-return as signal) is clever. One function with a clear flow is clear.

> *Ousterhout — deep modules over shallow ones.* `convertGatewayError` and `reclassifyError` are shallow modules. Collapsing them into `ConvertError` makes one deeper module with a simpler surface area for callers.

> *Effective Go — interfaces.* `ConvertError` is the `ErrorConverter` interface method. Making it the single authoritative path for error conversion means the interface contract and the internal behaviour are the same thing.

**Missing tests**: `convertGatewayError` has no direct tests. `reclassifyError` has no direct tests. Both are exercised indirectly through `TestPlatformModeErrorConversion`. Whether the code is restructured or not, error conversion needs a comprehensive table-driven test covering:

| Input error                                    | Status code | Expected output                           |
| ---------------------------------------------- | ----------- | ----------------------------------------- |
| raw `*openai.Error`                            | 402         | `InsufficientFundsError`                  |
| raw `*openai.Error`                            | 502         | `UpstreamProviderError`                   |
| raw `*openai.Error`                            | 504         | `GatewayTimeoutError`                     |
| raw `*openai.Error`                            | 401         | `AuthenticationError` (delegated to base) |
| raw `*openai.Error`                            | 429         | `RateLimitError` (delegated to base)      |
| already-converted `ProviderError` wrapping 502 | n/a         | `UpstreamProviderError` (reclassified)    |
| already-converted `AuthenticationError`        | n/a         | passed through unchanged                  |
| non-API error (network error)                  | n/a         | `ProviderError` (delegated to base)       |
| nil                                            | n/a         | nil                                       |

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+	case http.StatusGatewayTimeout:
+		return errors.NewGatewayTimeoutError(providerName, apiErr)
+	default:
+		return nil
+	}
+}
+
+// newNonPlatformProvider creates a gateway provider in non-platform mode.
+// The gateway API key is sent via the X-AnyLLM-Key header.
+func newNonPlatformProvider(cfg *config.Config, baseURL string) (*Provider, error) {
```
peteski22: `newNonPlatformProvider`/`newPlatformProvider` split.

These two functions share most of their structure: both build a `CompatibleConfig` with nearly identical fields, call `NewCompatible`, and return `&Provider{...}`. The differences are small: which auth mechanism to configure.

Parameter naming is poor: `newPlatformProvider(cfg *config.Config, baseURL, token string)` — `cfg` already has a `BaseURL` field so `baseURL` is redundant, and `token` is actually an API key being used as Bearer auth. Calling it `token` obscures what it is.

If the constructor is simplified per PR review comment about two-stage config parsing, then these two functions likely collapse. The mode difference is a few lines of branching, not two separate constructor functions.

> *Ousterhout — deep modules.* The split creates two parallel code paths that must be kept in sync. Every shared concern (capabilities, provider name, config fields) is duplicated. One path with a branch point is simpler to maintain.

> *Go Proverbs — "A little copying is better than a little dependency."* The two functions copy most of each other, but the duplication doesn't serve independence — they're tightly coupled to the same `CompatibleConfig` shape. The copying adds maintenance burden without benefit.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+			return converted
+		}
+	}
+
+	return err
+}
+
+// headerTransport wraps an http.RoundTripper to inject custom headers into
+// every request. Used in non-platform mode to add the X-AnyLLM-Key header.
+type headerTransport struct {
```
peteski22: `headerTransport` misplaced in file

The project's `CLAUDE.md` specifies file ordering:

1. Package declaration & imports
2. Constants
3. Interface assertions
4. Types (exported first, then unexported helpers)
5. Constructor (`New()`)
6. Exported methods (alphabetically)
7. Unexported methods (alphabetically)
8. Package-level functions (alphabetically)

`headerTransport` is an unexported type defined at line 206, sitting between unexported methods and package-level functions. It should be in the types section near line 48 with the `Provider` type.

`RoundTrip` is an exported method (required to satisfy `http.RoundTripper`) without a godoc comment. Per Effective Go, exported methods should have doc comments even on unexported types, because they're still visible through the interface.

**Missing tests**: `headerTransport.RoundTrip` has no direct unit test. It's exercised indirectly through `TestNonPlatformModeSendsCustomHeader`. A table-driven test should cover:

| Scenario                           | Expected                                  |
| ---------------------------------- | ----------------------------------------- |
| Base transport succeeds            | Header injected, response returned        |
| Base transport returns error       | Error propagated, no panic                |
| Request already has the header set | Header overwritten (Set vs Add semantics) |

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+	"github.com/mozilla-ai/any-llm-go/providers"
+	openaiProvider "github.com/mozilla-ai/any-llm-go/providers/openai"
+)
+
+// Provider configuration constants.
+const (
+	envAPIBase       = "GATEWAY_API_BASE"
+	envAPIKey        = "GATEWAY_API_KEY"
+	envPlatformToken = "GATEWAY_PLATFORM_TOKEN"
+	gatewayHeader    = "X-AnyLLM-Key"
```
peteski22: `X-AnyLLM-Key` literal repeated in comments

The header string is correctly extracted to the `gatewayHeader` constant. The literal string "X-AnyLLM-Key" then appears 7 more times across the two files, almost entirely in comments. If the header name changes, you'd update the constant but have 7 stale references in comments.

This is relevant now: [mozilla-ai/any-llm#1023](https://github.com/mozilla-ai/any-llm/issues/1023) suggests `AnyLLM-Key` as the canonical header, deprecating the `X-` prefix per [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648). Later, to stay relevant, the Go provider will need to update. Comments that reference the constant name rather than the literal value won't need touching.

> *Effective Go — commentary.* Comments should explain why, not restate what. "Inject the gateway authentication header" is more durable than "inject the X-AnyLLM-Key header."

Also note: `TestPlatformModeSendsBearerAuth` at line 344 uses the string literal in an assertion message (`"platform mode should not send X-AnyLLM-Key header"`) rather than referencing the constant, which is inconsistent with `TestNonPlatformModeSendsCustomHeader` which uses `gatewayHeader` correctly.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+			Transport: &headerTransport{
+				base:   baseTransport,
+				header: gatewayHeader,
+				value:  "Bearer " + gatewayKey,
+			},
+		}
+		compatOpts = append(compatOpts, config.WithHTTPClient(httpClient))
+	}
+
+	base, err := openaiProvider.NewCompatible(openaiProvider.CompatibleConfig{
```
peteski22: Explicit zero-value fields on `CompatibleConfig`

Both constructors explicitly set Go zero values:

```go
openaiProvider.NewCompatible(openaiProvider.CompatibleConfig{
    APIKeyEnvVar:   "",
    BaseURLEnvVar:  "",
    DefaultAPIKey:  "gateway-no-key",
    DefaultBaseURL: "",
    Name:           providerName,
    RequireAPIKey:  false,
    ...
})
```

Compare with `llamafile`, which explains intent:

```go
APIKeyEnvVar:   "", // Llamafile doesn't use an API key env var.
```

Without comments, a reader can't tell whether an empty string is deliberate ("this provider doesn't use an env var") or an oversight ("I forgot to fill this in"). The `BaseURLEnvVar: ""` is a good example: is it empty because the gateway resolves it manually, or because the author didn't realise it could be set to `envAPIBase`? (t's the latter, which a comment would have forced the author to articulate.)

> *Effective Go — commentary.* Comments should convey information not apparent from the code. An empty string is ambiguous — a comment resolves the ambiguity between "deliberately unused" and "accidentally omitted."

Note: The CLAUDE.md rule that requires setting all CompatibleConfig fields explicitly (including zero values) contradicts Go conventions. See: [#83](https://github.com/mozilla-ai/any-llm-go/issues/83) for the proposed revision.

providers/gateway/gateway.go
peteski22: `WithGatewayKey` stores via `config.WithExtra("gateway_key", key)`, then `resolveGatewayKey` retrieves via `cfg.ExtraValue("gateway_key")`. Same for `WithPlatformMode` / `"platform_mode"`.

The `config.Extra` map uses `any` values, so there's no type safety on the read side. `resolveGatewayKey` does `v.(string)` and `resolvePlatformMode` does `v.(bool)`, both with silent fallthrough on type mismatch. A bug like `WithExtra("platform_mode", "true")` (string instead of bool) would silently disable platform mode.

At minimum, the string keys should be named constants. But the broader concern is that `config.Extra` provides no type safety at all for provider-specific options.

> *Go Proverbs — "Errors are values."* The type assertion failures here are silent. A mismatched type returns the zero value and falls through to default behaviour — the error is swallowed rather than surfaced.

> *Ousterhout — "Define errors out of existence."* Named constants eliminate the class of typo bugs entirely. If the rule is that `WithExtra` keys must be constants, the compiler enforces coordination.

**Missing tests**: There's no test for type-mismatch fallthrough. A table-driven test should cover:

| Scenario                             | Expected                                   |
| ------------------------------------ | ------------------------------------------ |
| `WithExtra("gateway_key", "valid")`  | key resolved correctly                     |
| `WithExtra("gateway_key", 123)`      | wrong type — should error or be documented |
| `WithExtra("gateway_key", "")`       | empty string — should be treated as unset  |
| `WithExtra("platform_mode", true)`   | mode enabled                               |
| `WithExtra("platform_mode", "true")` | wrong type — currently silently ignored    |

These tests would document the actual contract and provide back-pressure to make the `config.Extra` mechanism more robust.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+		return true, platformToken
+	}
+
+	return false, ""
+}
+
+// capabilities returns the full set of capabilities for the gateway provider.
+// The gateway can proxy to any provider, so all features are marked as
+// supported. Actual support depends on the underlying provider being called.
+func capabilities() providers.Capabilities {
```
peteski22: `Capabilities` all-true is misleading for consumers

**Current behaviour**: The gateway returns all capabilities as `true`:

```go
func capabilities() providers.Capabilities {
    return providers.Capabilities{
        Completion:          true,
        CompletionImage:     true,
        CompletionPDF:       true,
        // ... all true
    }
}
```

**Problem**: A consumer calling `provider.Capabilities().Embedding` to decide whether to attempt an embedding request gets `true`, attempts it, and gets a runtime error from a backend that doesn't support embeddings. The `Capabilities` interface promises a contract these proxy providers can't honour.

**Context**: The platform provider does the same thing, so there is precedent. But precedent for a bad pattern doesn't justify propagating it. If the gateway is replacing the platform provider, this is the opportunity to improve the pattern rather than carry it forward.

**Root cause**: The `Capabilities` abstraction assumes static, known-at-construction-time capability sets, which doesn't fit proxy providers. This is a design limitation in the abstraction itself.

**Affected code**: `providers/gateway/gateway.go` lines 367-378, and `providers/platform/platform.go` lines 135-147 (same pattern).

> *Go Proverbs — "Don't just check errors, handle them gracefully."* If a consumer checks capabilities to avoid errors, returning `true` for everything undermines that error-avoidance path.

**Missing tests**: `TestCapabilities` verifies all fields are true (line 128-144), which confirms the current intent. But there's no test that verifies what error surfaces when a capability isn't actually supported by the upstream — e.g., calling `Embedding` through a gateway whose backend doesn't support it.

providers/gateway/gateway_test.go
peteski22, line 1: The PR extracts multiple small functions: `resolvePlatformMode`, `resolveGatewayKey`, `convertGatewayError`, `reclassifyError`, `forwardConfigOptions`; and the implicit justification for extracting them would be testability. But the tests don't follow through:

- `resolvePlatformMode`: 5 subtests, not table-driven, covers main paths but not edge cases
- `resolveGatewayKey`: 3 subtests, not table-driven
- `convertGatewayError`: no direct tests
- `reclassifyError`: no direct tests
- `forwardConfigOptions`: no direct tests (this is where the HTTPClient bug lives undetected)

Whether the code stays decomposed or is simplified per the feedback above, the input space is the same and needs the same systematic coverage. The shape of the tests might change, but the discipline doesn't.

> *Ousterhout — "Tactical vs strategic programming."* Extracting functions without writing the tests that justify the extraction is tactical — it looks like structure but doesn't deliver the reliability benefit. If a function is worth extracting, it's worth a table-driven test suite that covers the input space systematically.

> *Go Proverbs — "Don't just check errors, handle them gracefully."* Several of these functions have silent fallthrough on unexpected inputs (wrong type in Extra, nil config, empty strings). Tests should document what actually happens in those cases.

Specific test matrices for each area are included in the relevant PR feedback comments.

providers/gateway/gateway.go
```diff
@@ -0,0 +1,378 @@
...
+
+	chunks := make(chan providers.ChatCompletionChunk)
+	errs := make(chan error, 1)
+
+	go func() {
+		defer close(chunks)
+		defer close(errs)
+
+		for chunk := range upstreamChunks {
+			select {
```
peteski22, line 295: Streaming wrapper silently swallows errors on context cancellation

**Current behaviour**: In `CompletionStream` (line 136), if the context is cancelled mid-stream, the goroutine returns via the `case <-ctx.Done()` branch without draining `upstreamErrs`. The error channel closes empty, so a consumer reading `err := <-errs` gets `nil` and concludes the stream completed successfully.

**Expected behaviour**: The consumer should be able to distinguish between "stream completed successfully" and "stream was interrupted by context cancellation." At minimum, `context.Canceled` or `ctx.Err()` should surface on the error channel.

**Scope**: This is a **pre-existing pattern** — `CompatibleProvider.CompletionStream` (in `providers/openai/compatible.go` lines 167-204) does the same thing. The gateway adds a second channel-forwarding layer, which means there are now two places where context cancellation can silently swallow errors instead of one.

**Affected code**:

- `providers/gateway/gateway.go` lines 144-162 (gateway wrapper)
- `providers/openai/compatible.go` lines 174-204 (base provider — same pattern)

> *Go Proverbs — "Don't just check errors, handle them gracefully."* Silently dropping errors is the opposite of graceful handling. There's a reasonable justification (the caller cancelled, so they may not care about the error), but that justification should be documented and the behaviour should be deliberate rather than accidental.

> *Effective Go — concurrency.* The goroutine's cleanup path should be deliberate. Even if the decision is "do nothing with the error," that decision should be visible.

**Missing tests**: There is no test for context cancellation mid-stream. A test that cancels the context after receiving some chunks should verify:

1. the goroutine doesn't leak
2. the channels close cleanly,
3. the error channel surfaces an appropriate error (or the nil-on-cancel behaviour is documented as intentional via the test).

errors/errors.go
```diff
@@ -17,6 +17,9 @@ const (
 	CodeMissingAPIKey       = "missing_api_key"
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
+	CodeInsufficientFunds   = "insufficient_funds"
```
peteski22, line 20: Gateway-specific error types belong in the gateway package, not the shared errors package

Three new error types were added to `errors/errors.go`:

- `InsufficientFundsError` / `ErrInsufficientFunds` (HTTP 402)
- `UpstreamProviderError` / `ErrUpstreamProvider` (HTTP 502)
- `GatewayTimeoutError` / `ErrGatewayTimeout` (HTTP 504)

Plus three error code constants, three constructor functions, and re-exports in `anyllm.go`.

The existing errors in the package are provider-agnostic: `ErrRateLimit`, `ErrAuthentication`, `ErrModelNotFound`, `ErrContextLength`. Any provider could produce these. They describe *what went wrong* without assuming anything about architecture.

The new errors are different. `"Upstream provider"` and `"gateway timeout"` encode proxy architecture into their names. If Groq returned a 502, it would fall through to `ErrProvider`; and that's correct, because from the caller's perspective Groq isn't "upstream" of anything. But now `ErrUpstreamProvider` exists in the shared package, creating an asymmetry: a 502 from the gateway gets a specific type, while a 502 from Groq gets the generic one.

`ErrInsufficientFunds` (402) is arguably more generic ; any paid API could return that, and it has a reasonable case for living in the shared package. But `ErrUpstreamProvider` and `ErrGatewayTimeout` are gateway domain concepts.

These should be defined in `providers/gateway/` as package-local error types. They can still embed `errors.BaseError` and work with `errors.As`. If consumers need `errors.Is` matching, the gateway package exports its own sentinels. This keeps the shared errors package focused on universally applicable categories and avoids polluting `anyllm.go` with gateway-specific re-exports.

> *Go Proverbs — "The bigger the interface, the weaker the abstraction."* Applied to the error package as a shared abstraction boundary. Every type added is a commitment that all consumers of any-llm-go must understand. Gateway-specific types weaken that abstraction by including concepts relevant to only one provider.

> *Ousterhout — information leakage.* Gateway implementation details (the concept of "upstream providers," the proxy architecture) leak into a shared module. Consumers who never use the gateway see these types in their IDE autocomplete and documentation.

> *Effective Go — package design.* Packages should provide a coherent set of related types. The errors package currently does this well — everything is about "what went wrong talking to an LLM provider." Adding "what went wrong inside a proxy" muddies that cohesion.

The error type tests in `errors/errors_test.go` are well-written and should move with the types if they're relocated. The gateway's `TestPlatformModeErrorConversion` and `TestNonPlatformModeAlsoConvertsGatewayErrors` test the end-to-end conversion path and would remain unchanged regardless of where the types live.

peteski22 (changes requested):
For context: The `gateway` provider is intended to replace `providers/platform`, but neither the PR description nor the code acknowledges this relationship. Since nothing in main is tagged or released, there's no backwards compatibility constraint.

This PR should either:

1. Remove the platform provider and position gateway as its replacement, or
2. `git mv providers/platform providers/gateway` first, then refactor; preserving git history and making intent clear in the diff.

Shipping both side-by-side with no documentation about which to use creates confusion for anyone looking at the codebase. The PR description should be explicit about the strategic intent, and the package doc comment should reference the relationship.

This context also matters because several design choices in this PR replicate patterns from the platform provider (capabilities all-true, `WithExtra` for options). If the gateway is replacing the platform provider, this is the opportunity to improve those patterns, not carry them forward.

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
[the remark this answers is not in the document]
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

providers/gateway/gateway_test.go
tbille: Done

providers/gateway/gateway.go
tbille: Done

errors/errors.go
tbille: Done

peteski22 (approved):
👍🏼

PR #89, opened 2026-04-15
feat: add batch API support with gateway provider

tbille:
## Summary

Add BatchProvider interface and a new Gateway provider that implements both completions and batch operations against the any-llm gateway HTTP API.

## Changes

- Add batch types to `providers/types.go`: `Batch`, `BatchStatus`, `BatchRequestCounts`, `BatchRequestItem`, `CreateBatchParams`, `ListBatchesOptions`, `BatchResult`, `BatchResultItem`, `BatchResultError`
- Add `BatchProvider` interface following the optional-interface pattern of `EmbeddingProvider`
- Add `BatchNotCompleteError` and `ErrBatchNotComplete` sentinel to `errors/errors.go`
- Re-export all batch types, status constants, and error types from root `anyllm.go`
- Create `providers/gateway/` package with full `Provider` implementation:
  - `Completion()` and `CompletionStream()` (SSE) for chat completions
  - `CreateBatch()`, `RetrieveBatch()`, `CancelBatch()`, `ListBatches()`, `RetrieveBatchResults()` for batch operations
  - `doRequest()` helper with `X-AnyLLM-Key: Bearer <key>` auth header
  - HTTP error mapping: 401/403 → AuthenticationError, 404 → ProviderError with upgrade hint, 409 → BatchNotCompleteError, 422 → ProviderError, 429 → RateLimitError, 502 → ProviderError
  - `parseBatchNotCompleteDetail()` regex parser for extracting batch ID and status from 409 error messages
- Add `"gateway"` to `testutil` provider env key map
- Comprehensive unit tests in `providers/gateway/gateway_test.go` (933 lines):
  - Construction: `TestNewRequiresAPIBase`, `TestNewWithAPIKey`, `TestNewWithoutAPIKey`, `TestNewFromEnvVars`
  - Capabilities verification
  - Completion: success, auth header, HTTP error mapping, streaming (SSE parsing, context cancellation, error propagation)
  - Batch: create, auth header, retrieve with provider param, cancel, list with pagination, retrieve results
  - Error mapping: 409 → BatchNotCompleteError (with `errors.Is` check), 404 → upgrade hint, 401 → AuthenticationError, 422 → ProviderError, 502 → upstream error
  - `convertParamsToRequest` coverage for all CompletionParams fields
  - `parseBatchNotCompleteDetail` regex edge cases

## Testing

- All unit tests pass: `go test ./... -race -count=1`
- All tests use `httptest` fake servers with `t.Parallel()` and `require` assertions
- Integration tests gated by `testutil.SkipIfNoAPIKey("gateway")`
- No existing tests affected

## Checklist

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,7 +18,6 @@ const (
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
 	CodeInsufficientFunds   = "insufficient_funds"
-	CodeBatchNotComplete    = "batch_not_complete"
 )
 
 // Sentinel errors for type checking with errors.Is().
@@ -34,7 +33,6 @@ var (
 	ErrUnsupportedProvider = stderrors.New("unsupported provider")
 	ErrUnsupportedParam    = stderrors.New("unsupported parameter")
 	ErrInsufficientFunds   = stderrors.New("insufficient funds")
-	ErrBatchNotComplete    = stderrors.New("batch not yet complete")
 )
 
 // BaseError is the base error type for all any-llm errors.
@@ -295,24 +293,4 @@ func NewInsufficientFundsError(provider string, err error) *InsufficientFundsErr
 	}
 }
 
-// BatchNotCompleteError is returned when retrieve_batch_results is called
-// on a batch that is not yet complete.
-type BatchNotCompleteError struct {
-	BaseError
-	BatchID string
-	Status  string
-}
 
-// NewBatchNotCompleteError creates a new BatchNotCompleteError.
-func NewBatchNotCompleteError(provider string, batchID string, status string) *BatchNotCompleteError {
-	return &BatchNotCompleteError{
-		BaseError: BaseError{
-			Code:     CodeBatchNotComplete,
-			Provider: provider,
-			Err:      fmt.Errorf("batch '%s' is not yet complete (status: %s)", batchID, status),
-			sentinel: ErrBatchNotComplete,
-		},
-		BatchID: batchID,
-		Status:  status,
-	}
-}
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -292,5 +292,3 @@ func NewInsufficientFundsError(provider string, err error) *InsufficientFundsErr
 		},
 	}
 }
-
-
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,6 +18,7 @@ const (
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
 	CodeInsufficientFunds   = "insufficient_funds"
+	CodeBatchNotComplete    = "batch_not_complete"
 )
 
 // Sentinel errors for type checking with errors.Is().
@@ -33,6 +34,7 @@ var (
 	ErrUnsupportedProvider = stderrors.New("unsupported provider")
 	ErrUnsupportedParam    = stderrors.New("unsupported parameter")
 	ErrInsufficientFunds   = stderrors.New("insufficient funds")
+	ErrBatchNotComplete    = stderrors.New("batch not yet complete")
 )
 
 // BaseError is the base error type for all any-llm errors.
@@ -290,5 +292,27 @@ func NewInsufficientFundsError(provider string, err error) *InsufficientFundsErr
 			Err:      err,
 			sentinel: ErrInsufficientFunds,
 		},
+	}
+}
+
+// BatchNotCompleteError is returned when retrieve_batch_results is called
+// on a batch that is not yet complete.
+type BatchNotCompleteError struct {
+	BaseError
+	BatchID string
+	Status  string
+}
+
+// NewBatchNotCompleteError creates a new BatchNotCompleteError.
+func NewBatchNotCompleteError(provider string, batchID string, status string) *BatchNotCompleteError {
+	return &BatchNotCompleteError{
+		BaseError: BaseError{
+			Code:     CodeBatchNotComplete,
+			Provider: provider,
+			Err:      fmt.Errorf("batch '%s' is not yet complete (status: %s)", batchID, status),
+			sentinel: ErrBatchNotComplete,
+		},
+		BatchID: batchID,
+		Status:  status,
 	}
 }
--- a/providers/gateway/gateway.go
+++ b/providers/gateway/gateway.go
@@ -19,7 +19,6 @@ import (
 	stderrors "errors"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -634,12 +633,12 @@ func (p *Provider) callBatchJSON(
 	return nil
 }
 
-// closeBody closes an HTTP response body and logs any close error; intended
-// for use in defer statements.
+// closeBody closes an HTTP response body; intended for use in defer
+// statements. The close error is deliberately discarded because this SDK
+// does not expose a logger through config, and callers have already
+// consumed the body by the time this runs.
 func closeBody(resp *http.Response) {
-	if err := resp.Body.Close(); err != nil {
-		log.Printf("gateway: failed to close response body: %v", err)
-	}
+	_ = resp.Body.Close()
 }
 
 // doRequest sends an HTTP request to the gateway API for batch operations.
--- a/providers/gateway/gateway.go
+++ b/providers/gateway/gateway.go
@@ -259,19 +259,28 @@ func New(opts ...config.Option) (*Provider, error) {
 		)
 	}
 
-	// Resolve apiBase for batch API calls.
+	// Resolve apiBase once here and pass the resolved value through to
+	// NewCompatible via WithBaseURL. Resolving in a single place avoids
+	// the possibility of the batch HTTP client and the OpenAI SDK client
+	// disagreeing (e.g. if config state changed between resolutions) and
+	// lets us surface resolution errors directly instead of relying on
+	// NewCompatible's RequireBaseURL check.
 	apiBase, err := cfg.ResolveBaseURL(envAPIBase, "")
-	if err != nil || apiBase == "" {
-		// Will be caught by NewCompatible's RequireBaseURL, but resolve
-		// for our own use.
-		apiBase = ""
+	if err != nil {
+		return nil, err
+	}
+	if apiBase == "" {
+		return nil, fmt.Errorf(
+			"gateway base URL is required (set via WithBaseURL option or %s env var)",
+			envAPIBase,
+		)
 	}
 	apiBase = strings.TrimRight(apiBase, "/")
 
 	// Pass the user's opts straight through and layer auth-specific opts on
-	// top (order matters: later options win). Base URL resolution and the
-	// required-URL check are delegated to NewCompatible via BaseURLEnvVar
-	// and RequireBaseURL.
+	// top (order matters: later options win). WithBaseURL is appended last
+	// so NewCompatible sees the already-resolved, trimmed URL and does not
+	// re-run ResolveBaseURL.
 	compatOpts := slices.Clone(opts)
 	if platformMode {
 		compatOpts = append(compatOpts, config.WithAPIKey(platformToken))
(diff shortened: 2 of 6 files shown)


providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+
+	var detail struct {
+		Detail string `json:"detail"`
+	}
+	_ = json.Unmarshal(bodyBytes, &detail)
+
+	msg := detail.Detail
+	if msg == "" {
+		msg = string(bodyBytes)
+	}
```
peteski22: ❓ does the gateway return structured JSON responses? If so that would be better for us, and we wouldn't need a brittle regex.

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+
+	switch resp.StatusCode {
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return errors.NewAuthenticationError(providerName, fmt.Errorf("%s", msg))
+	case http.StatusNotFound:
+		if strings.Contains(path, "/v1/batches") {
+			return errors.NewProviderError(providerName,
+				fmt.Errorf("this gateway does not support batch operations; upgrade your gateway"))
```
peteski22: Do we ever hit this, don't all the path's have `/v1/batches` in?

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+		return nil, err
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Printf("gateway: failed to close response body: %v", err)
+		}
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, p.handleBatchError(resp, "/v1/batches")
```
peteski22: Should extract this to a const to prevent magic strings appearing throughout the code.

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+		return errors.NewRateLimitError(providerName, fmt.Errorf("%s", msg))
+	case http.StatusBadGateway:
+		return errors.NewProviderError(providerName, fmt.Errorf("upstream provider error: %s", msg))
+	default:
+		return errors.NewProviderError(providerName, fmt.Errorf("HTTP %d: %s", resp.StatusCode, msg))
+	}
+}
+
+// batchNotCompleteRE matches: Batch 'batch_abc' is not yet complete (status: in_progress)
+var batchNotCompleteRE = regexp.MustCompile(`[Bb]atch\s+'([^']+)'.*\(status:\s*(\w+)\)`)
```
peteski22: File ordering will have this near the top (if it's still required).

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+
+// parseBatchNotCompleteDetail extracts batch ID and status from the error detail.
+func parseBatchNotCompleteDetail(detail string) (batchID, status string) {
+	matches := batchNotCompleteRE.FindStringSubmatch(detail)
+	if len(matches) >= 3 {
+		return matches[1], matches[2]
+	}
+	return "", "unknown"
+}
```
peteski22: This is a brittle approach.

errors/errors.go
```diff
@@ -18,6 +18,7 @@ const (
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
 	CodeInsufficientFunds   = "insufficient_funds"
+	CodeBatchNotComplete    = "batch_not_complete"
```
peteski22: Are these errors only specific to the gatway? If so, let's move them to the gateway package for now.

errors/errors.go
```diff
@@ -292,3 +294,25 @@ func NewInsufficientFundsError(provider string, err error) *InsufficientFundsErr
...
 	}
 }
+
+// BatchNotCompleteError is returned when retrieve_batch_results is called
+// on a batch that is not yet complete.
+type BatchNotCompleteError struct {
+	BaseError
+	BatchID string
+	Status  string
+}
```
peteski22: In wrong place in the file, but also might need to be moved to gateway only (even unexported).

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+	_ = json.Unmarshal(bodyBytes, &detail)
+
+	msg := detail.Detail
+	if msg == "" {
+		msg = string(bodyBytes)
+	}
+
+	switch resp.StatusCode {
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return errors.NewAuthenticationError(providerName, fmt.Errorf("%s", msg))
```
peteski22: If you just have a string and nothing else to add, just use `stderrors.New(msg)`, but be aware that in Go you should not capitalise the first letter of the error msg... so might be better to add some context first.

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+		return nil, errors.NewInvalidRequestError(providerName, err)
+	}
+
+	resp, err := p.doRequest(ctx, http.MethodPost, "/v1/batches", body)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Printf("gateway: failed to close response body: %v", err)
```
peteski22: As this is an SDK we aren't currently (I hope) logging anywhere else in the library, we'd have to accept a logger through config and wire it through providers if we want/need this.

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+	if err != nil {
+		return nil, errors.NewProviderError(providerName, err)
+	}
+
+	return resp, nil
+}
+
+// handleBatchError handles HTTP error responses and maps them to typed errors.
+func (p *Provider) handleBatchError(resp *http.Response, path string) error {
+	bodyBytes, _ := io.ReadAll(resp.Body)
```
peteski22, line 697: Could do with a comment to explain why you're ignoring the error here.

tbille:
review the code

providers/gateway/gateway.go
```diff
@@ -395,3 +435,262 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+	return &batch, nil
+}
+
+// RetrieveBatch retrieves a batch job by ID.
+func (p *Provider) RetrieveBatch(
+	ctx context.Context,
+	batchID string,
+	provider string,
+) (*providers.Batch, error) {
+	path := fmt.Sprintf("/v1/batches/%s?provider=%s", url.PathEscape(batchID), url.QueryEscape(provider))
```
peteski22:
```go
u := &url.URL{
      Path: "/v1/batches/" + batchID, // or path.Join if you prefer
      RawQuery: url.Values{"provider": {provider}}.Encode(),
}
fullURL := p.apiBase + u.String()
```

providers/gateway/gateway.go
```diff
@@ -192,6 +212,15 @@ func New(opts ...config.Option) (*Provider, error) {
...
 
+	// Resolve apiBase for batch API calls.
+	apiBase, err := cfg.ResolveBaseURL(envAPIBase, "")
+	if err != nil || apiBase == "" {
+		// Will be caught by NewCompatible's RequireBaseURL, but resolve
+		// for our own use.
+		apiBase = ""
+	}
+	apiBase = strings.TrimRight(apiBase, "/")
+
```
peteski22, line 287: The base URL is being resolved twice: once here for the batch HTTP client, once inside NewCompatible for the OpenAI SDK client. At (1), the error is ignored because the assumption is "(2) will fail loudly if resolution really didn't work, so I don't need to error here."

Two problems with that:

  - If resolution succeeds in cfg.ResolveBaseURL but returns different output from the resolution NewCompatible does internally (e.g. because cfg state changed, or an option is order-sensitive), the provider ends up with a batch URL and a completions URL that disagree. Unlikely today, but brittle.
  - If resolution fails here, apiBase = "" is stored on the Provider struct — and then NewCompatible errors before New returns, so the empty apiBase never gets used in practice. The dead assignment makes the code harder to read.

Constructive fix: resolve once, surface the error, pass the resolved URL through so NewCompatible doesn't re-resolve:

```go
apiBase, err := cfg.ResolveBaseURL(envAPIBase, "")
  if err != nil {
      return nil, err
  }
  if apiBase == "" {
      return nil, fmt.Errorf(
          "gateway base URL is required (set via WithBaseURL or %s)",
          envAPIBase,
      )
  }
  apiBase = strings.TrimRight(apiBase, "/")

  compatOpts = append(compatOpts, config.WithBaseURL(apiBase))
  // NewCompatible now uses the already-resolved URL; no re-resolution, no double-resolution risk.
```

peteski22 (changes requested):
Left some feedback to resolve. <3

providers/gateway/gateway.go
```diff
@@ -395,3 +500,269 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+	if msg == "" {
+		return nil
+	}
+	return stderrors.New(msg)
+}
+
+// decodeBatchErrorBody parses a gateway error response. Non-JSON bodies are
+// preserved via the raw field so callers can still surface the server's
+// text in the typed error.
+func decodeBatchErrorBody(body []byte) batchErrorBody {
```
peteski22: Two concerns:

**Name mismatches behavior.** `decode` implies strict parse; the function silently swallows `json.Unmarshal` errors and falls back to raw bytes. If the gateway ever renames `detail` → `message`, or a proxy returns HTML, the code keeps appearing to work — no log, no metric, no test signal. Same silent-breakage class as the original regex.

**Body size (minor, future).** `io.ReadAll(resp.Body)` reads without bound, so a misbehaving upstream could in theory balloon the error string. Unlikely in practice today, but worth keeping in mind if you ever see odd payloads here.

Idiomatic fix — surface the parse error so drift is visible:

```go
type batchError struct {
    Detail, Status string
}

func parseBatchError(r io.Reader) (batchError, error) {
    body, err := io.ReadAll(r)
    if err != nil {
        return batchError{}, err
    }
    var b batchError
    if err := json.Unmarshal(body, &b); err != nil {
        return batchError{Detail: string(body)}, err
    }
    return b, nil
}
```

`handleBatchError` then decides per-status whether a parse failure is tolerable (log and fall through for generic errors; fail hard for 409 where `Status` matters).

providers/gateway/gateway.go
```diff
@@ -395,3 +500,269 @@ func newHeaderClient(base *http.Client, headerValue string) *http.Client {
...
+	if err := p.callBatchJSON(ctx, http.MethodGet, path, batchID, nil, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
+}
+
+// batchItemPath builds a path for /v1/batches/{id}[/action]?provider=X using
+// url.URL so that batchID and provider are encoded safely.
+func batchItemPath(batchID, action, provider string) string {
```
peteski22, line 606: The `url.URL{Path, RawQuery}` shape matches my earlier suggestion, but the switch dropped the `url.PathEscape(batchID)` that the original code had, and I don't think that was intentional.

`url.URL.EscapedPath()` escapes `?`, `#`, space, etc., but **not `/`** — it's a path separator in the decoded `Path` field. So a `batchID` containing `/` or `..` would traverse into a sibling route rather than being encoded as a single segment. Practically batch IDs are probably `batch_<alnum>`, but strictly this is weaker than the pre-review code.

Cleaner idiomatic fix using `url.URL.JoinPath` (Go 1.19+), which escapes segments for you:

```go
func batchItemPath(batchID, action, provider string) string {
    u := (&url.URL{Path: batchesPath}).JoinPath(batchID)
    if action != "" {
        u = u.JoinPath(action)
    }
    u.RawQuery = url.Values{providerQueryParam: {provider}}.Encode()
    return u.RequestURI()
}
```

Keeps composition flat and handles escaping for both `batchID` and `action` without manual `PathEscape` calls.

providers/gateway/gateway.go
tbille: Addressed in 5c189e8604.

providers/gateway/gateway.go
tbille: Addressed in 5c189e8604.

peteski22 (approved):

PR #92, opened 2026-04-17
Add moderation API support to gateway provider

tbille:
## Summary

Adds content moderation support to the Go SDK's gateway provider, exposing the gateway's `POST /v1/moderations` endpoint through a new `ModerationProvider` interface. Introduces `*errors.UnsupportedError` so callers can cleanly detect backends (e.g. Anthropic) that do not offer moderation.

## Changes

- Add `ModerationParams`, `ModerationResult`, `ModerationResponse`, `ContentPart`, `ImageURL` and the optional `ModerationProvider` interface in `providers/types.go`; add `Moderation bool` to `providers.Capabilities`.
- Add `UnsupportedError`, `ErrUnsupported` sentinel, `CodeUnsupported` constant, and `NewUnsupportedError` constructor in `errors/errors.go` to distinguish "provider does not support this operation" from the existing "provider not registered" case.
- Re-export the new types, interface, and sentinel from the root `anyllm` package to match the existing alias pattern.
- Implement `Moderation` on `*gateway.Provider` by calling `POST /v1/moderations` directly with the captured `*http.Client`, so `IncludeRaw` can be forwarded as `?include_raw=true` and the gateway's locked `"does not support moderation"` / `"multimodal moderation"` 400 phrasing can be mapped to `*UnsupportedError`. Capture `baseURL`, `httpClient` and the platform token on `Provider` at construction so direct HTTP calls share the same auth as the embedded openai-go SDK. Advertise `Moderation: true` in `capabilities()` and add the `providers.ModerationProvider` interface assertion.
- Add `httptest`-based tests covering the happy path, the `include_raw` query parameter, both unsupported phrasings (moderation / multimodal) mapping to `*UnsupportedError`, status-code mapping for 401/402/429/502/504, platform and non-platform auth on the direct HTTP call, and an env-gated integration test.
- Add a README snippet showing the gateway provider's `Moderation` method and how to branch on `UnsupportedError`.

## Testing

- `go test ./errors ./providers/gateway -race -count=1` (targeted to packages changed)
- Integration test `TestIntegrationModeration` is env-gated on `GATEWAY_API_BASE` / `GATEWAY_PLATFORM_TOKEN` and skips when unset.
- Full suite runs in CI.

## Checklist

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,6 +18,7 @@ const (
 	CodeUnsupportedProvider = "unsupported_provider"
 	CodeUnsupportedParam    = "unsupported_parameter"
 	CodeInsufficientFunds   = "insufficient_funds"
+	CodeUnsupported         = "unsupported"
 )
 
 // Sentinel errors for type checking with errors.Is().
@@ -33,6 +34,7 @@ var (
 	ErrUnsupportedProvider = stderrors.New("unsupported provider")
 	ErrUnsupportedParam    = stderrors.New("unsupported parameter")
 	ErrInsufficientFunds   = stderrors.New("insufficient funds")
+	ErrUnsupported         = stderrors.New("operation not supported by provider")
 )
 
 // BaseError is the base error type for all any-llm errors.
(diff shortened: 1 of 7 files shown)


errors/errors.go
```diff
@@ -156,6 +158,22 @@ type InsufficientFundsError struct {
...
+// UnsupportedError indicates a provider does not support a given operation
+// (e.g. moderation). Use errors.As to detect:
+//
+//	var unsup *errors.UnsupportedError
+//	if errors.As(err, &unsup) {
+//	    // unsup.Operation, unsup.Provider ...
+//	}
+//
+// Also matches errors.Is(err, ErrUnsupported).
+type UnsupportedError struct {
```
peteski22:
```suggestion
type UnsupportedOperationError struct {
```

peteski22 (approved):

PR #95, opened 2026-04-22
feat(anthropic): add json_schema structured output support

channel-dante:
## Summary

- Adds `json_schema` structured output support to the Anthropic provider by mapping `CompletionParams.ResponseFormat` to `anthropic.MessageNewParams.OutputConfig.Format` via the stable (non-beta) Anthropic SDK API
- Extracts an `applyResponseFormat` helper following the same pattern as `applyThinking` and sibling providers (OpenAI, Gemini, Ollama, Deepseek)
- Adds `responseFormatJSONObject` / `responseFormatJSONSchema` constants matching the naming convention in other providers

## Behavior

| Input | Result |
|---|---|
| `ResponseFormat = nil` | No change (existing behavior) |
| `Type = "json_object"` | No-op — Anthropic requires a schema; documented in code |
| `Type = "json_schema"`, `JSONSchema = nil` | No-op |
| `Type = "json_schema"`, valid schema | `OutputConfig.Format.Schema` set |

`Name`, `Description`, and `Strict` fields from `JSONSchema` are not forwarded — `anthropic.JSONOutputFormatParam` only accepts `Schema` and `Type`.

The feature works for both `Completion` and `CompletionStream` via the shared `convertParams`.

## Test Plan

- [x] 4 unit tests covering all input cases (`TestConvertParams_ResponseFormat`)
- [x] Integration test skipped without `ANTHROPIC_API_KEY` (`TestIntegrationCompletionWithStructuredOutput`)
- [x] Full test suite (`go test ./...`) passes with no regressions

### DIFF
--- a/providers/anthropic/anthropic.go
+++ b/providers/anthropic/anthropic.go
@@ -213,15 +213,7 @@ func (p *Provider) convertParams(params providers.CompletionParams) (anthropic.M
 		req.ToolChoice = convertToolChoice(params.ToolChoice, params.ParallelToolCalls)
 	}
 
-	if params.ResponseFormat != nil &&
-		params.ResponseFormat.Type == responseFormatJSONSchema &&
-		params.ResponseFormat.JSONSchema != nil {
-		req.OutputConfig = anthropic.OutputConfigParam{
-			Format: anthropic.JSONOutputFormatParam{
-				Schema: params.ResponseFormat.JSONSchema.Schema,
-			},
-		}
-	}
+	applyResponseFormat(&req, params.ResponseFormat)
 
 	applyThinking(&req, params.ReasoningEffort, maxTokens)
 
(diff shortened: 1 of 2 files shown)


providers/anthropic/anthropic.go
```diff
@@ -390,6 +398,26 @@ func applyThinking(req *anthropic.MessageNewParams, effort providers.ReasoningEf
 	}
 }
 
+// applyResponseFormat configures structured output on the request if applicable.
+func applyResponseFormat(req *anthropic.MessageNewParams, format *providers.ResponseFormat) {
```
peteski22, line 402: Could we add a subtest under `TestConvertParams_ResponseFormat` that passes a `providers.JSONSchema` with `Name`, `Description`, and `Strict` set alongside `Schema`, and asserts the result has only `OutputConfig.Format.Schema` populated? 

The PR body and inline comment both note that those three fields are dropped because the Anthropic SDK's `JSONOutputFormatParam` has no destination for them; pinning that in a test would make the contract explicit and guard against future SDK changes that accidentally pick one up.

peteski22 (commented):
The PR notes that structured output works for both `Completion` and `CompletionStream` via the shared `convertParams`. That's structurally true, but the unit tests only exercise the non-streaming path and the integration test calls `Completion`, not `CompletionStream`. 

Could we add one assertion that a `params` struct destined for the streaming path also produces the expected `OutputConfig.Format.Schema`? 

A call to `p.convertParams(params)` with a valid `json_schema` format, asserting the `OutputConfig` is set, is enough... we don't need to spin up a real stream.

peteski22 (approved):
🙏🏼 thanks @channel-dante. LGTM.

PR #141, opened 2026-08-27
fix(openai): align current Chat controls

IceCodeNew:
## Summary

- Migrate the existing Chat and embedding consumers from `openai-go` v1.12.0 to `openai-go/v3` v3.42.0.
- Honor `OPENAI_BASE_URL` without overriding an explicit `WithBaseURL`.
- Preserve the active `max_completion_tokens` request key and every explicit reasoning effort, including `none`; omit only the binding's `auto` sentinel.
- Expose and preserve omitted, false, and true function `strict` values for OpenAI-compatible providers.
- Keep DeepSeek and Mistral token-key differences inside their provider transforms.

This PR does not add Responses, Files, Batch, media APIs, embedding input variants, shared `Extra` policy, or previously unsupported deprecated request parameters.

## SDK and documentation baseline

- Current Chat reference: <https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create>
- Current GPT-5.6 guide: <https://developers.openai.com/api/docs/guides/latest-model>
- Current DeepSeek strict-tool guide: <https://api-docs.deepseek.com/guides/tool_calls>
- Latest official Go SDK audited: [`v3.56.0`](https://github.com/openai/openai-go/tree/f5b985771236464300abc9395c1c4abda0d65c53)
- Minimum retained dependency: [`v3.42.0`](https://github.com/openai/openai-go/tree/86bbd3d91826b36b3bd88460869d87584f11eed7), the first checked formal release whose typed `ReasoningEffort` catalog includes all seven current Chat values

The retained SDK already exposes `FunctionDefinitionParam.Strict` and distinguishes omitted from explicit false, so strict tool serialization requires no dependency upgrade. DeepSeek strict mode additionally requires callers to select its documented `/beta` base URL and set every function strict; this binding does not silently reroute endpoints or duplicate server-side JSON Schema validation.

`openai-go` uses open string-backed unions, so model-specific reasoning support remains service-side. The binding does not infer that every model accepts every value.

## Review boundary

The superseded head changed normalized `MaxTokens` from the binding's existing `max_completion_tokens` wire behavior to deprecated `max_tokens`. This head restores `max_completion_tokens`; no deprecated request surface is introduced.

Ablation removed the shared `Extra` allowlist and tests, the deleted DeepSeek/Mistral transforms, the deprecated OpenAI token path, and a standalone formatting commit. For strict tools, deleting the three-line converter mapping makes both explicit false and true disappear while omission remains unchanged, so the public field and mapping are both required.

The owning diff is 11 files with +215/-83 lines across five signed commits. The strict-tool change is one 48-line commit including its wire regression. Most of the remaining changes adapt generated v3 tagged unions and their existing tests. No OpenAI SDK or Fantasy code or fixtures were copied.

## Exact-head evidence

Base: `ef366a4218ed67ad82156742c985829fc389df3d`

Head: `a8fbfeee85de77133e967bd894c690138aac4079`

```text
go test ./... -count=1                                                               PASS
CGO_ENABLED=1 go test -race ./providers/openai ./providers/deepseek \
  ./providers/mistral ./providers/groq ./providers/gateway -count=1                 PASS
go test with temporary openai-go v3.56.0 modfile for the same providers             PASS
go mod tidy -diff                                                                     no diff
go mod verify                                                                         all modules verified
golangci-lint run --new-from-rev=d6327122b119c1854a48f399956c68fa84354512 \
  ./providers/...                                                                     0 issues
git diff --check                                                                      PASS
```

The new strict-tool wire test verifies omitted, explicit false, and explicit true independently. Latest-SDK source inspection confirms v3.56.0 retains the same `param.Opt[bool]` omission contract. A temporary modfile execution against v3.56.0 passes for OpenAI, DeepSeek, Mistral, Groq, and Gateway.

Live OpenAI and DeepSeek strict-tool verification and exact-head CodeRabbit review remain pending. This PR remains Draft and does not claim complete provider alignment.

AI tools were used for contract research, implementation, and local verification. Claims above were checked against current official documentation and immutable SDK source.

### DIFF
--- a/providers/openai/openai.go
+++ b/providers/openai/openai.go
@@ -9,6 +9,7 @@ import (
 const (
 	defaultBaseURL = "https://api.openai.com/v1"
 	envAPIKey      = "OPENAI_API_KEY"
+	envBaseURL     = "OPENAI_BASE_URL"
 	providerName   = "openai"
 )
 
@@ -31,6 +32,7 @@ type Provider struct {
 func New(opts ...config.Option) (*Provider, error) {
 	base, err := NewCompatible(CompatibleConfig{
 		APIKeyEnvVar:   envAPIKey,
+		BaseURLEnvVar:  envBaseURL,
 		Capabilities:   capabilities(),
 		DefaultBaseURL: defaultBaseURL,
 		Name:           providerName,
--- a/anyllm.go
+++ b/anyllm.go
@@ -53,11 +53,14 @@ const (
 
 // ReasoningEffort levels.
 const (
-	ReasoningEffortAuto   = providers.ReasoningEffortAuto
-	ReasoningEffortHigh   = providers.ReasoningEffortHigh
-	ReasoningEffortLow    = providers.ReasoningEffortLow
-	ReasoningEffortMedium = providers.ReasoningEffortMedium
-	ReasoningEffortNone   = providers.ReasoningEffortNone
+	ReasoningEffortAuto    = providers.ReasoningEffortAuto
+	ReasoningEffortHigh    = providers.ReasoningEffortHigh
+	ReasoningEffortLow     = providers.ReasoningEffortLow
+	ReasoningEffortMax     = providers.ReasoningEffortMax
+	ReasoningEffortMedium  = providers.ReasoningEffortMedium
+	ReasoningEffortMinimal = providers.ReasoningEffortMinimal
+	ReasoningEffortNone    = providers.ReasoningEffortNone
+	ReasoningEffortXHigh   = providers.ReasoningEffortXHigh
 )
 
 // Provider types.
@@ -179,17 +182,17 @@ var (
 
 // Error types.
 type (
-	AuthenticationError           = errors.AuthenticationError
-	BaseError                     = errors.BaseError
-	ContentFilterError            = errors.ContentFilterError
-	ContextLengthError            = errors.ContextLengthError
-	InsufficientFundsError        = errors.InsufficientFundsError
-	InvalidRequestError           = errors.InvalidRequestError
-	MissingAPIKeyError            = errors.MissingAPIKeyError
-	ModelNotFoundError            = errors.ModelNotFoundError
-	ProviderError                 = errors.ProviderError
-	RateLimitError                = errors.RateLimitError
-	UnsupportedOperationError     = errors.UnsupportedOperationError
-	UnsupportedParamError         = errors.UnsupportedParamError
-	UnsupportedProviderError      = errors.UnsupportedProviderError
+	AuthenticationError       = errors.AuthenticationError
+	BaseError                 = errors.BaseError
+	ContentFilterError        = errors.ContentFilterError
+	ContextLengthError        = errors.ContextLengthError
+	InsufficientFundsError    = errors.InsufficientFundsError
+	InvalidRequestError       = errors.InvalidRequestError
+	MissingAPIKeyError        = errors.MissingAPIKeyError
+	ModelNotFoundError        = errors.ModelNotFoundError
+	ProviderError             = errors.ProviderError
+	RateLimitError            = errors.RateLimitError
+	UnsupportedOperationError = errors.UnsupportedOperationError
+	UnsupportedParamError     = errors.UnsupportedParamError
+	UnsupportedProviderError  = errors.UnsupportedProviderError
 )
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/mozilla-ai/any-llm-platform-client-go v0.0.1
 	github.com/ollama/ollama v0.32.5
-	github.com/openai/openai-go v1.12.0
+	github.com/openai/openai-go/v3 v3.42.0
 	github.com/stretchr/testify v1.11.1
 	google.golang.org/genai v1.66.0
 )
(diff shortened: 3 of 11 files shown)


providers/openai/openai.go
```diff
@@ -54,5 +57,7 @@ func capabilities() providers.Capabilities {
 		CompletionTools:     true,
 		Embedding:           true,
 		ListModels:          true,
+		Moderation:          true,
```
XYenon: `Moderation: true` looks inconsistent with the provider interfaces in this PR.

`openai.Provider` (and the embedded `CompatibleProvider`) still does not implement `providers.ModerationProvider` / `Moderation(...)`, while `Capabilities().Moderation` is changed to `true`. A caller using capabilities for feature discovery can therefore see moderation support and then fail the interface assertion / have no moderation method to call.

XYenon (commented):
I compared this PR with the current Python `mozilla-ai/any-llm` Responses implementation, and there are a few Go-specific parity issues worth addressing:

1. `ResponsesResult` collapses the full OpenAI Response into only `ID`, `Model`, and `OutputText()`. This loses non-text output items such as function calls and reasoning items, plus response status/incomplete details. The PR itself notes that tool-call and reasoning-only turns can legitimately have empty `OutputText`, but with the current wrapper those successful turns become indistinguishable from an actually empty response. The Python implementation returns the full `ResponseResource` / native `Response` instead of collapsing it to text.

2. `ResponsesParams.Reasoning` reuses the chat-completion `ReasoningEffort` abstraction. Python keeps Responses `reasoning` as the native reasoning object and forwards it to the Responses API. In the Go implementation, `none` is omitted entirely while `auto` is accepted and forwarded, which does not match the Responses API semantics. It would be better either to mirror Python's native `reasoning` structure or at least map `auto` to omission and preserve `none` explicitly where supported.

3. The new Responses API surface is much narrower than Python's: input is limited to `{role, content string}` messages, so prior response output/reasoning items cannot be replayed, and tools/tool_choice/previous_response_id/etc. are unavailable. This may be acceptable for an initial implementation, but the response-side lossiness in (1) makes tool-call turns unusable rather than merely unsupported.

Also, as noted separately, `Moderation: true` is currently advertised without implementing `ModerationProvider`, while Python's OpenAI provider has both the capability and the implementation.

providers/openai/openai.go
IceCodeNew: Correction after the rebuild: the current head no longer advertises or implements Moderation. #141 now contains only the minimum OpenAI Go SDK v3 migration, so the capability/interface inconsistency reported here is absent. My previous reply described a superseded head.

IceCodeNew:
Correction for the rewritten current head: this comment described a superseded Responses and Moderation implementation. PR #141 no longer advertises or implements either API. It now contains only the OpenAI SDK v3 migration for existing Chat and embedding consumers, the OPENAI_BASE_URL fix, and two Chat request-presence fixes.

The current head preserves explicit reasoning_effort values such as none and omits only the binding auto sentinel. The official current Chat reference lists none, minimal, low, medium, high, xhigh, and max, with model-dependent support: https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create. The GPT-5.6 guide lists none, low, medium, high, xhigh, and max for that model family: https://developers.openai.com/api/docs/guides/latest-model.

The prior statements about ResponsesResult, native Responses methods, and Moderation are not claims about the current PR.

XYenon:
> **2. `ResponsesParams.Reasoning` reusing `ReasoningEffort`.** The Responses API `reasoning.effort` accepts only `low`/`medium`/`high`. Both `auto` and `none` are mapped to omission (the model uses its default), with an explicit comment explaining the mapping. Unsupported efforts (`minimal`, `xhigh`, `max`) return `UnsupportedParamError` via `supportedResponsesReasoning()`. This keeps the unified `ReasoningEffort` abstraction while respecting the Responses API's narrower accepted set.

https://developers.openai.com/api/docs/guides/latest-model?utm_source=chatgpt.com

> Set reasoning.effort intentionally. GPT-5.6 supports none, low, medium, high, xhigh, and max.
> If you are migrating from GPT-5.5 or GPT-5.4, preserve your current reasoning effort as the baseline, then compare one level lower.
> If you use none, keep it as your latency baseline and also test low when the workflow benefits from reasoning or tool use.
> Use medium as a balanced starting point and low for latency-sensitive workloads.
> Use high or xhigh when more reasoning produces a measured quality gain.
> Reserve max for the hardest quality-first workloads. Compare max and xhigh to find the best quality, latency, and cost tradeoff for your use case.

IceCodeNew:
Correction: my previous conclusion was wrong. I treated the generated constants in the repository's old `openai-go` release as a closed enum. `shared.ReasoningEffort` is string-backed, so those constants did not prevent the SDK from sending newer documented values. Current `openai-go` also declares `none`, `minimal`, `xhigh`, and `max` explicitly.

OpenAI's current GPT-5.6 guide documents `none`, `low`, `medium`, `high`, `xhigh`, and `max`: https://developers.openai.com/api/docs/guides/latest-model. The PR now forwards the official SDK effort values, including explicit `none`, `xhigh`, and `max`, while treating `auto` as omission.

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
Rebased all five signed commits onto current upstream `ef366a4218ed67ad82156742c985829fc389df3d`. Exact head is `b71a5aeb8ffcdf4febffd05da7539656287fef6b`; full tests, targeted race tests, module tidy/verify, and GolangCI-Lint 2.13.2 (`./providers/...`, 0 issues) pass.

@coderabbitai full review

IceCodeNew:
Correction to the preceding local-lint report: the repository GolangCI configuration enables formatter fixes, so the earlier `0 issues` run had changed two Gateway lines in the worktree without committing them. Exact head `e0e908b37160dc02f9258eb0b80cee5d125d4c71` now includes that formatter output as a separate signed `+4/-2` commit. `golangci-lint run --fix=false ./providers/...` reports `0 issues` with a clean worktree, and the full tests, targeted race tests, module tidy/verify, and diff check pass.

@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
IceCodeNew consolidated this work into https://github.com/mozilla-ai/any-llm-go/pull/176, https://github.com/mozilla-ai/any-llm-go/pull/177. The replacement PRs target mozilla-ai/main and retain the accepted implementation and tests in clean per-layer commits. Dependent layers stay draft until their predecessors merge and the resulting upstream diff is revalidated. Closing this superseded PR to avoid duplicate review; the original branch, commits and local audit evidence are preserved. Prepared by Amp for IceCodeNew.

PR #140, opened 2026-08-27
fix(gemini): align current thinking controls

IceCodeNew:
## Summary

- use `thinkingLevel` for Gemini 3 and documented token budgets for Gemini 2.5 and Robotics
- preserve provider defaults for `auto`, send an explicit zero where thinking can be disabled, and reject unsupported combinations before transport
- expose the normalized `minimal` effort and verify the generated SDK request with an independent HTTP wire test

## Contract sources

- [Gemini thinking guide](https://ai.google.dev/gemini-api/docs/generate-content/thinking), updated 2026-08-17
- [go-genai v1.71.0 `ThinkingConfig`](https://github.com/googleapis/go-genai/blob/v1.71.0/types.go#L2735-L2744)
- [go-genai v1.71.0 Gemini request converter](https://github.com/googleapis/go-genai/blob/v1.71.0/models.go#L1295-L1298)

The current dependency remains `google.golang.org/genai v1.56.0`. All four current thinking levels entered the SDK in v1.40.0, so this change does not require a dependency upgrade. A temporary modfile test against the latest formal release, v1.71.0, also passes.

## Review boundary

This PR contains five files and three signed commits. Its diff is `+368/-125`: production `+190/-68`, direct contract tests `+178/-57`. It does not contain Interactions, Batch, embeddings, file/PDF handling, base URL changes, response capture, or general stream rewrites.

The model service validates each Gemini 3 model's supported subset of `minimal`, `low`, `medium`, and `high`. Keeping that catalog in the binding would duplicate server validation and become stale when Google adds models.

## Verification

Base: upstream `ef366a4218ed67ad82156742c985829fc389df3d`. Exact head: `4d0ef23ffd6ef8def7c63759da3157f45ab2c467`.

```text
go test ./... -count=1                                      PASS
CGO_ENABLED=1 go test -race ./... -count=1                  PASS
go mod tidy -diff                                            no diff
go mod verify                                                all modules verified
golangci-lint run --fix=false . ./providers/gemini/...      0 issues
git diff --check                                            PASS
```

`golangci-lint run --fix=false ./providers/gemini/...` reports `0 issues.`. The full repository still reports two formatter findings in unchanged `providers/gateway/gateway.go`; #141 owns those fixes because it modifies that shared SDK consumer. They remain absent from this standalone Gemini PR.

The generated SDK HTTP golden captures a real `GenerateContent` request and checks the documented `generationConfig.thinkingConfig` wire object. No upstream code or fixture was copied.

[...]

PR #142, opened 2026-08-27
feat(azureopenai): add the current v1 API-key provider

IceCodeNew:
## Summary

Add an Azure OpenAI provider for the current `/openai/v1/` API-key contract. Azure owns endpoint and authentication policy; the shared OpenAI-compatible adapter retains Chat, stream, embeddings, model listing, context, and error behavior.

This PR remains a draft. Real Azure service verification and exact-head CodeRabbit review are pending. Entra ID, preview media, Responses, Files, and Batch are separate responsibilities.

## Scope

- Resolve `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT`, with explicit options taking precedence.
- Send API keys in Azure's documented `api-key` header and remove ambient OpenAI bearer authentication.
- Use `{endpoint}/openai/v1/` without a dated `api-version`; treat `model` as the Azure deployment name.
- Reuse shared OpenAI Chat messages, streaming, tools, reasoning, embeddings, model listing, context, and normalized errors.
- Require HTTPS and reject endpoint credentials, query parameters, fragments, authentication-header overrides, and unsupported `WithExtra` values.
- Keep credentials on the configured endpoint origin through openai-go's native HTTP client guard.

## Boundary and ablation

The earlier implementation added a local credential transport and direct transport tests. openai-go v3.53.0 owns this security boundary for native `*http.Client` requests and redirects. This rebuild deletes the local transport, origin helper, and 45-line direct transport test. A 307 redirect contract test exercises `http.Client.Do` and confirms the second origin receives no request.

`CompatibleConfig.ClientOptions` remains as a small shared seam because Azure must configure a base URL, HTTP client, and authentication headers without copying Chat, stream, embeddings, models, context, and error lifecycle code. `OpenAIMessageSchema` is explicit because the generic compatible path deliberately exposes a narrower schema; Azure v1 uses the current OpenAI message schema.

The owning range is seven signed commits after #145 exact head `8e95c9a29734ad039b8ba9b5ff8b7c2091f15109`, across nine files at `+571/-37`. Production code adds 155 lines, direct feature tests add 333 lines, docs add 28 lines, and the remaining changes are module metadata. The tests contain no copied upstream fixture or provider-owned protocol lifecycle.

## Dependency

This PR depends on #145 exact head `8e95c9a29734ad039b8ba9b5ff8b7c2091f15109`, including its #141 dependency.

It upgrades `github.com/openai/openai-go/v3` from v3.42.0 to v3.53.0. v3.53.0 is the first formal release with generic configured-origin checks for native HTTP clients. An otherwise identical temporary build at v3.52.0 failed `TestAzureV1RejectsCredentialRedirect` because the redirected origin received one request. v3.53.0 rejected the redirect before transport. The current v3.55.0 source and behavior were also checked.

The SDK's `azure.WithAPIKey` cannot replace this construction: it requires `azure.WithEndpoint`, which still adds a dated query and legacy `/openai/deployments/{deployment}` route. Microsoft v1 requires `/openai/v1/` and puts the deployment name in `model`.

## Sources

- Microsoft v1 lifecycle: <https://learn.microsoft.com/azure/foundry/openai/api-version-lifecycle>
- Microsoft v1 REST reference: <https://learn.microsoft.com/azure/foundry/openai/latest>
- Azure v1 OpenAPI at commit `9bf1953`: <https://github.com/Azure/azure-rest-api-specs/blob/9bf19532443271e5c02c7968bea48ba632b749ad/specification/ai/data-plane/OpenAI.v1/azure-v1-v1-generated.json>
- OpenAI Chat create reference: <https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create>
- openai-go v3.55.0 Azure adapter: <https://github.com/openai/openai-go/blob/v3.55.0/azure/azure.go>
- openai-go v3.53.0 HTTP client contract: <https://github.com/openai/openai-go/blob/v3.53.0/option/requestoption.go#L42-L59>
- Fantasy architecture reference: <https://github.com/charmbracelet/fantasy/tree/73eb01dec54ee7923b9c77d3e03cc9fd78b28cf9/providers>

Fantasy was used only to review the provider boundary. No Fantasy code, control flow, fixture, test, comment, or assertion sequence was copied. Fantasy and openai-go are Apache-2.0; this independently written PR redistributes neither source.

## Verification

- `go test ./... -count=1`: passed.
- `CGO_ENABLED=1 go test -race ./providers/azureopenai ./providers/openai -count=1`: passed.
- `go mod tidy -diff`: no diff.
- `go mod verify`: `all modules verified`.
- `golangci-lint run --fix=false ./...`: `0 issues`.
- `golangci-lint fmt --diff ./...`: no diff.
- `git diff --check`: passed.
- All seven owning commits pass local SSH signature verification.

All 112 nondeprecated GolangCI-Lint rules were run on the owning range. Actionable `noinlineerr`, `wsl_v5`, and `goconst` reports were fixed. The remaining 31 reports consist of no-policy `depguard` imports, demands for global sentinels for ordinary constructor failures, exhaustive initialization of optional request and capability structs, one cohesive multi-endpoint wire test over the linter's 60-line threshold, and white-box tests that exercise private constructor policy. No suppression or lint configuration was added.

The tests use independently written literal HTTP requests and responses. They cover route and query serialization, the Azure v1 `developer` role and participant `name`, explicit `reasoning_effort: "none"`, API-key isolation, unknown response fields, Chat, stream terminal, embeddings, models, a 401 Azure error envelope, unsupported extras, and cross-origin redirect rejection.

## Incomplete

- No Azure credential was available. Real Chat, streaming, embeddings, models, deployment and region behavior, service redirects, throttling, cancellation, and timeout remain unverified.
- Azure-specific live cancellation and timeout remain unverified; local evidence comes from the shared compatible lifecycle used without overrides.
- Entra ID, preview media, Responses, Files, and Batch are not implemented or advertised by this PR.
- Exact-head CodeRabbit review is pending after this rewrite.


XYenon:
A few issues I noticed while reviewing this as a self-contained PR:

1. **Some docs appear to include changes from #140/#141.** For example, `docs/providers.md` says Gemini 3+ uses `thinking_level` and that OpenAI implements `ResponsesProvider`, but neither feature exists on this PR's `main` base. Since the PR description says this branch is self-contained on `main`, I think those unrelated doc changes should be removed.

2. **Azure's Responses capability will become inconsistent once #141 lands.** `azureopenai.Provider` embeds `*openai.Provider`, so after #141 it will automatically implement `providers.ResponsesProvider`, but `azureopenai.capabilities()` still leaves `Responses` false. That would give:
   - `provider.(providers.ResponsesProvider)` → supported
   - `provider.Capabilities().Responses` → false

   It would be good to coordinate this with #141 so the interface and capability flag stay consistent.

3. **`ListModels` reuses the OpenAI model converter, but Azure's model-list response has a different metadata shape.** The Azure REST API uses fields such as `created_at`, while `openai-go`'s `Model` expects `created` and `owned_by`. As a result, `CompatibleProvider.ListModels()` is likely to return `Created == 0` and an empty `OwnedBy` for real Azure responses. A fixture using the actual Azure response shape, or an Azure-specific model converter, would catch this.

The core `azure.WithEndpoint` / `azure.WithAPIKey` client setup itself looks good.

IceCodeNew:
Correction for the rebuilt PR at `c9ac2886cb77e2f44e4e52f881dfda36aa5edc17`:

1. The provider documentation is now Azure-specific.
2. Azure embeds `openai.CompatibleProvider`, which does not implement `providers.ResponsesProvider`. This PR therefore keeps `Capabilities().Responses == false`; Azure Responses is outside this API-key core slice and needs its own independently reviewable PR.
3. My earlier `created_at` statement was wrong for the current v1 contract. The current Azure v1 OpenAPI model schema uses `created`, so the shared OpenAI model converter is correct. The rebuilt wire fixture verifies `created` mapping and tolerance of unknown Azure metadata: https://github.com/Azure/azure-rest-api-specs/blob/a6943a926f76b3a2f90371b3466157eddf760e25/specification/ai/data-plane/OpenAI.v1/azure-v1-v1-generated.json

No Azure-specific model converter remains in this PR.

XYenon:
One behavioral difference from the Python `any-llm` `AzureopenaiProvider` worth flagging, in case it's unintentional: this provider does not expose any `api-version` control for chat completions.

In this PR, chat completions (along with embeddings, models, Files, and Batch) route through the GA `{endpoint}/openai/v1/` surface with no `api-version` query — `v1Options`/`v1BaseURL` never set one. Only the media client (`mediaClientOptions`) pins `option.WithQuery("api-version", "preview")` for image/speech/transcription. There's also no `api_version` config option or extra key consumed anywhere.

The Python `AzureopenaiProvider` behaves differently — its `_init_client` reads an `api_version` (kwarg first, else `OPENAI_API_VERSION` env, default `"preview"`) and forwards it to `AsyncAzureOpenAI(api_version=...)`, which applies the `api-version` query to all requests including chat:

```python
api_version = kwargs.pop("api_version", None) or os.getenv("OPENAI_API_VERSION", self.DEFAULT_API_VERSION)
...
self.client = AsyncAzureOpenAI(api_key=api_key, azure_endpoint=azure_endpoint, api_version=api_version, ...)
```

So a consumer migrating from the Python library — or one passing an `api_version` extra by analogy — can pin a specific dated api-version for chat there, but here that value has no effect and is silently dropped (`WithExtra` doesn't error on unknown keys). Is dropping per-request api-version selection for chat (in favor of the GA v1 route) an intentional divergence? If so, it'd be worth documenting, since it's a silent no-op rather than a startup error for anyone relying on it.

IceCodeNew:
> One behavioral difference from the Python `any-llm` `AzureopenaiProvider` worth flagging, in case it's unintentional: this provider does not expose any `api-version` control for chat completions.
> 
> In this PR, chat completions (along with embeddings, models, Files, and Batch) route through the GA `{endpoint}/openai/v1/` surface with no `api-version` query — `v1Options`/`v1BaseURL` never set one. Only the media client (`mediaClientOptions`) pins `option.WithQuery("api-version", "preview")` for image/speech/transcription. There's also no `api_version` config option or extra key consumed anywhere.
> 
> The Python `AzureopenaiProvider` behaves differently — its `_init_client` reads an `api_version` (kwarg first, else `OPENAI_API_VERSION` env, default `"preview"`) and forwards it to `AsyncAzureOpenAI(api_version=...)`, which applies the `api-version` query to all requests including chat:
> 
> ```python
> api_version = kwargs.pop("api_version", None) or os.getenv("OPENAI_API_VERSION", self.DEFAULT_API_VERSION)
> ...
> self.client = AsyncAzureOpenAI(api_key=api_key, azure_endpoint=azure_endpoint, api_version=api_version, ...)
> ```
> 
> So a consumer migrating from the Python library — or one passing an `api_version` extra by analogy — can pin a specific dated api-version for chat there, but here that value has no effect and is silently dropped (`WithExtra` doesn't error on unknown keys). Is dropping per-request api-version selection for chat (in favor of the GA v1 route) an intentional divergence? If so, it'd be worth documenting, since it's a silent no-op rather than a startup error for anyone relying on it.

Thanks for bringing it up. I'm still working on this PR, that will be fixed later.
The GPT-5.6 x-high really shitted a lot in these days...

IceCodeNew:
Correction after rebuilding #142 at `c9ac2886cb77e2f44e4e52f881dfda36aa5edc17`:

This feedback identified a real silent no-op. The rebuilt provider deliberately targets Microsoft’s current `/openai/v1/` API-key contract, which does not use a dated `api-version`: https://learn.microsoft.com/azure/foundry/openai/api-version-lifecycle

Because `api_version` is not part of that contract, the constructor now rejects every non-empty `config.WithExtra(...)` map, including `api_version`, instead of silently accepting it. `TestNewRejectsUnsupportedExtra` covers this startup error. Preview media and its separate `api-version=preview` contract are outside this PR.

My previous statement that `WithAPIVersion("v1"|"preview")` fixed the issue referred to an implementation that has since been removed. It was not the correct fix for the current v1 core provider.

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
@coderabbitai full review

IceCodeNew:
IceCodeNew consolidated this work into https://github.com/mozilla-ai/any-llm-go/pull/179. The replacement PRs target mozilla-ai/main and retain the accepted implementation and tests in clean per-layer commits. Dependent layers stay draft until their predecessors merge and the resulting upstream diff is revalidated. Closing this superseded PR to avoid duplicate review; the original branch, commits and local audit evidence are preserved. Prepared by Amp for IceCodeNew.

